PR #15, opened 2017-03-23
fix #14 (ugly way)

### LINKED ISSUE #14: There's two buttons for show logo under Misc.
Yeah. They sync up to each other but I have no idea why there are two. Fuckin weird shit.

no-boot-device:
This isn't the best looking way to fix it. In my opinion, having ADDCVARS take a string with wildcards as opposed to a string just to check for occurrences would definitely be better. However, the best method would be changing CatVar so that all categorization is done by the file initializing it. That would take a lot of rewriting, but if it were implemented correctly, could clean up the code considerably and make adding new hacks easier without reducing speed too much.


nullifiedcat:
fixed by my own commit

PR #63, opened 2017-04-18
just some file structure documentation

no-boot-device:
("""documentation""")


nullifiedcat:
hacks/chams is going to be removed sooner or later, no need to put anything in that place

PR #115, opened 2017-05-10
Added ubercharge esp

julianacat:


julianacat:
The requested changes were made + I saved cpu cycles and prevented a possible conflicting var for the body aim detection

PR #117, opened 2017-05-11
use sudo with gdb

no-boot-device:
we already run `sudo ./detach` so i have no clue why we don't `sudo gdb` :thinking:


ricefrag:
When running a script with sudo, all commands within it are ran with sudo.

no-boot-device:
yes, but i mean that "we already ask for a password just to detach, why don't we let gdb have it so it can attach without the script being ran as sudo"

ricefrag:
Ah, gotcha

PR #120, opened 2017-05-11
ADD TRIGGETBOT KEY

iRapThaGod:
How about to add triggerbot key or chosing mode like always or key? or always for backstabs,vaccinate,airblast and key for shooting? or chossable?


ricefrag:
This isn't a place to request features.

PR #168, opened 2017-07-04
Added gcc6 installation script for Arch.

SylveonBottle:
With this added, users won't need to "find gcc6 on your own".


nullifiedcat:
Actually gcc6 installation and cathook dependencies installation should be different in both ubuntu and arch install script. Split scripts into 2 different gcc/deps installation scripts and I'll merge it

PR #170, opened 2017-07-04
Removed redundancies & user interaction.

SylveonBottle:
The current installation script requires 3 user interactions:

1. Password for sudo
2. base-devel "select packages"
3. gcc in conflict with gcc-mutilibs.

Fix for 1 could be to use yaourt instead. 
Fix for 2 would be to manually install the packages inside the group - but which ones are quired for cathook compilation?
Fix for 3 is to use --noconfirm.

On second thought, would base-devel even be required? It's basically the first thing ever installed on arch machines. And the conflic with gcc and gcc-multilibs is because base-devel installs gcc.
Removing base-devel would fix 2.

After testing the script, using --noconfirm to fix 3 auto-no's. But, using "yes | " before the script auto-yes's. 

tl;dr Removed base-devel from installation, since it's more than likely already installed. Auto-confirm for gcc-multilib in case in conflict with gcc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ sudo apt update && sudo apt install build-essential software-properties-common -
 
 Arch dependencies installation::
 ```bash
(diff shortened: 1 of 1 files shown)


SylveonBottle:
TODO: Install base-devel anyway, maybe we can use the same method for gcc-mutlilib auto-confirm to select all in base-devel? Will require more testing. 

```bash
echo 1-25 ^10 | sudo pacman -S base-devel gdb gdb-common glew1.10 lib32-glew1.10 --noconfirm
```

julianacat:
If you could, could you provide your discord id @SylveonBottle

SylveonBottle:
Provided in discord.

PR #179, opened 2017-07-05
Someone borked installation script

SylveonBottle:
bash update instead of bash update-data in cathook installation

Now that ./update also runs ./update-data


nullifiedcat:
update will just re-compile it again, there's no need for that after make

PR #187, opened 2017-07-09
.gitignore

polivilas:
nice folders in main branch dude


polivilas:
using it myself, ignores non-essential files

PR #190, opened 2017-07-13
Aimbot charge-waiting / Removecond pickup updates + Better Code of Conduct

Retarded-Skid:
### Aimbot Changes:

* Added CatVar for always firing at 100% regardless if it will kill. Off by default
    * Useful for Mann Vs. Machine larger robots that have 30K health

### Removecond Changes:

* Edited CatVar for health pickups to be more generic, added explicit support for ammo packs
    * Fixed Removecond activating if the pickup was not yet spawned
    * Changed value for picking up items to use the user's Removecond value instead of 950 *(Which caused me to lag out 4/5 times with 40ms ping)*
        * Hardcoding may work for one person, but letting the user figure it out is better for everyone

### Code of Conduct Changes:

* New one's TLDR: Anyone is accepted no matter what. Just be an adult.


julianacat:
Please check the codacy checks, the  wait for charge could be improved.   
As black fire said in the issue, the equasion could be wrong, but it should only shoot if it can kill as thats the point of wait for charge. If you want to shoot at full charge then manually do that yourself. Adding a bypass to shoot at 100% seems kinda useless to me.     
   
As for the remove cond, i dont entirely know if the ce_bad check is good or not. If the player needs to remove cond before the player reaches the pack, once the pack is picked up there could be a few ticks that the player isnt using remove cond, potentially preventing the remove cond from making it infinite as the player would pick it up before the remove cond comes into effect. I havent tested but I dont know, please tell me if it works. 
And with the entire auto-remove cond for pickups should be moved into the `if (!exticks)` and instead of setting the amount, it would just add exticks like the rest of the auto-remove cond things, though this isnt part of your commit it would be nice to see it moved up there but its nice to see you made the remove cond value used by it chosen by user settings.
As with making the ammo pickup check, you can check the users ammo-count by using netvars, there is a netvar for clip on the weapon entity, you can make a for loop to go through the players list of weapons and check if any are not full on ammo, otherwise i wouldnt make it remove cond for ammo packs as there is no telling if it is needed.
Final thing with the remove cond would be to set a remove cond value for each usecase, for example, if it needed to use infinite healthpacks it would use 900, while with instant rezoom it could use something else, preferably lower. As with you lagging out, a fix could be to scale the value by latency or to scale it with the user setting but currently it works fine with the user setting for everything.    
   
With my final complaint going to the conduct, why tho...
I get it, I like :b:emes, you like :b:emes, but seriously, please :b:ont :/   
     
     
As with all commits by people, the final say goes to cat, lets see what he thinks about this.

Retarded-Skid:
> As black fire said in the issue, the equasion could be wrong, but it should only shoot if it can kill as thats the point of wait for charge. If you want to shoot at full charge then manually do that yourself. Adding a bypass to shoot at 100% seems kinda useless to me.

Some people like automation. As [Slyveon stated](https://github.com/nullifiedcat/cathook/issues/188) its useful for Mann Vs Machine. For the people who want it it's there. For the people who don't its off by default.

> As for the remove cond, i dont entirely know if the ce_bad check is good or not. If the player needs to remove cond before the player reaches the pack, once the pack is picked up there could be a few ticks that the player isnt using remove cond, potentially preventing the remove cond from making it infinite

I added it because otherwise it would removecond on empty pickup spaces, which is pointless. I've tested it in various pubs and it seems to work flawlessly *(And not break it like you hypothesize)*. My ping has ranged between 20-400 in these pubs if that is relevant information.

> And with the entire auto-remove cond for pickups should be moved into the if (!exticks) and instead of setting the amount, it would just add exticks like the rest of the auto-remove cond things

I'll see if I can do that without breaking it.

> Final thing with the remove cond would be to set a remove cond value for each usecase,

May I suggest moving Removecond to its own tab so you can fit in more options? Some options would be case-specific values like you mentioned.

> Please check the codacy checks

Its all not having `{}` for in-lined if statements, which are already extremely prevalent in the Cathook code base. Inline ifs are much more compact and make code more legible in multiple cases. 

```cpp
if (e->m_iClassID != CL_CLASS(CBaseAnimating)) continue;
if (e->m_flDistance > 54.0f) continue;
```

Fixing them here is kinda dumb IMO, but if you insist I'll change them.

julianacat:
>Some people like automation. As Slyveon stated its useful for Mann Vs Machine. For the people who want it it's there. For the people who don't its off by default.
:thinking: 

>I added it because otherwise it would removecond on empty pickup spaces, which is pointless. I've tested it in various pubs and it seems to work flawlessly (And not break it like you hypothesize). My ping has ranged between 20-400 in these pubs if that is relevant information.
If it doesnt break, then I approve!

>I'll see if I can do that without breaking it.
I saw the commit and it looks good, it should work in that place, I dont know what cat things of it though...

>May I suggest moving Removecond to its own tab so you can fit in more options? Some options would be case-specific values like you mentioned.
True, Im just suggesting that because not everything has to be one value for it to work, but having an entire section dedicated to remove cond seems kinda weird to me. Looks good in the misc area like it is now.

>Its all not having {} for in-lined if statements, which are already extremely prevalent in the Cathook code base. Inline ifs are much more compact and make code more legible in multiple cases.
Codacy approved so it looks good to me. 

>I'll can undo the commit. But sensible adults don't need to be told not to be assholes.
understandable, have a nice day

Retarded-Skid:
>> And with the entire auto-remove cond for pickups should be moved into the if (!exticks) and instead of setting the amount, it would just add exticks like the rest of the auto-remove cond things

> I'll see if I can do that without breaking it.

Done. Still works as before the switch.

PR #207, opened 2017-07-21
Strings must use doublequote. (quotes)

SylveonBottle:
### Why is this an issue?
JavaScript allows you to define strings in one of three ways: double quotes, single quotes, and backticks (as of ECMAScript 6). Each of these lines creates a string and, in some cases, can be used interchangeably. The choice of how to define strings in a codebase is a stylistic one outside of template literals (which allow embedded of expressions to be interpreted). Many codebases require strings to be defined in a consistent manner. This rule takes as parameter the style you prefer ("double", "single" or "backtick") default is double.
```
//Bad with double:
var single = 'single';
var unescaped = 'a string containing "double" quotes';
```

Also I was gonna add codacy button but decided against it, then realised I have no idea how to un-commit shit so I just made more commits fixing it or something idk


SylveonBottle:
Codacy confirms 22 issues fixed with this pull request. Nothing too big or ever to likely break something, but it feels better having Codacy score this project a tad bit higher now.

I'll look into other small things I can fix up to make Codacy happy - just small things that even someone like me knows won't break shit.

PR #268, opened 2017-08-15
Updated cathook information for gcc-7 compatibility on arch

SylveonBottle:
I can confirm that cathook compiles 100% with no errors using `gcc (GCC) 7.1.1 20170630` with arch. gcc-6 is not present on my machine. I did not use the provided "cc=gcc-7" code. I have updated cathook readme.md to account for this new information, updating the "list of what cathook requires". I have kept gcc-6 as a requirement, as users have reported being unable to compile cathook without gcc-6, but only for Ubuntu users.

I've changed the arch dependencies script to use the latest gcc (gcc 7.1.1 as of this pull request), instead of gcc-6 since it's no longer required.
I also added glew to the dependencies list, as it's found in the installation script.

I have not touched Ubuntu dependencies/gcc-6 script, as gcc-6 is (reportedly) required, and I would be unable to test those changes.

I wouldn't be making this pull request if I wasn't confident in the changes I made.


SylveonBottle:
Closed in favor of #282

PR #282, opened 2017-08-16
Update and rename update-gccdowngrade to arch-gcc-fix

SylveonBottle:
https://puu.sh/xbGu8/36d0f01c87.png


SylveonBottle:
update-gccdowngrade has been replaced with arch-gcc-fix, which is a better alternative. This will install gcc6, backup the files, then upgrade to gcc7 and place the files back. This way, gcc7 will be installed, but cathook can still use gcc6.

This pull will replace the arch dependencies script with one that will basically run arch-gcc-fix.
arch-gcc-fix is redundant and can now be removed, but will be kept for now due to users who may rather run the script rather than re-run the arch dependencies script.

oneechanhax:
your push needs to be edited, cathook now automatically puts the cathook data folder in a global location and update-data isnt required anymore

PR #285, opened 2017-08-17
the makefile now works for 80% more pc's!

ItsiAdam:
you people are fucking retarded


oneechanhax:
What if someone has gcc multilib installed, how would they switch to the gcc6 compiler

ItsiAdam:
this needs a better solution overall, which searching for it isn't the best idea

PR #299, opened 2017-09-02
Adding preferred Targets to Followbot in roaming mode

BenCat07:
Can still be Optimized, but this works fine for now.


oneechanhax:
The first problem is that i see is that its hard to determine what changed as github or you did something to break it.     
Second is that you could've made the code simpler with a switch instead of those "ifs"
Next, instead of putting the preference code into the roaming target selector, you added a checker in front of it. This wont work well as the target selector can always select that target again, and again, and again after the thing above it attempted to change it.    
Put a check into the area where the selector gets its score, if the target is a preferred, than boost its score by alot.
Sorry, I don't like the code in this >_>

BenCat07:
@oneechanhax i'm very sorry, i accidentaly didn't upload the optimized code, got one much better now.

PR #312, opened 2017-12-09
fixed sdk submodule

hkva:


BenCat07:
[submodule "source-sdk-2013"]
 	path = source-sdk-2013
had to be changed to
[submodule "source-sdk-2013-headers"]
 	path = source-sdk-2013-headers

first.

PR #314, opened 2017-12-10
Fixed CanShoot reporting false on weapon reload

snipermain:
Should fix a big annoying bug which has existed for ages. where the aimbot will not shoot with silent aim when reloading because the nextprimaryattack netvar reports that the next attack is when the reload finishes, which is in the future. This can be remedied by only updating nextattack if we have fired the weapon. This works with 99% of weapons, except flamethrowers - for some reason this change makes canshoot always return true. You might also want to check for flamethrowers and use the traditional method for them. Also i'm not sure if this code even compiles since I don't have the bandwidth to clone your repo right now, but i'm sure if it doesn't it should be an easy fix.

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -901,12 +901,19 @@ bool CanHeadshot()
 
 bool CanShoot()
 {
-
-    float servertime, nextattack;
-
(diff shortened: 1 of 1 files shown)


thinkingmaster:
Stale

PR #378, opened 2018-04-13
b1g gayness removal

mrsteyk:
Apparently everyone has b1g gay for not doing that like fucking forever ago


BenCat07:
no need since it only dies on clang

mrsteyk:
its needed since it dies on clang. clang > all. It compiles a lot faster and has potential for doing some more cool things if shit hits the fan.

PR #426, opened 2018-05-08
Removed unused code.

SylveonBottle:
How'd that get there


SylveonBottle:
Used as a reference. 80% of this stuff _is_ used. Go through and delete the stuff I found that isn't, at least.

PR #441, opened 2018-05-16
Disable backtrack by default and add UI setting

TotallyNotElite:
I don't think backtrack should be enabled by default since it makes for a bad user experience in my opinion. For example: 
- It creates annoying dots that don't go away in some occasions
- It lags the game out if aimbot is used in conjunction with backtrack (?)
- There is no GUI option to disable it

### DIFF
--- a/data/menu.json
+++ b/data/menu.json
@@ -523,6 +523,13 @@
 			},
 			{
 				"type": "list",
+				"name": "Backtrack",
+				"list": [
+					"backtrack"
+				]
+			},
+			{
+				"type": "list",
 				"name": "Debug Settings",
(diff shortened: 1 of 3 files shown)


buyBread:
just add "aimbot_backtrack" into the menu ?

TotallyNotElite:
That's not the point I'm trying to make though. If you *need* backtrack you should be able to enable it. But most will be annoyed by the issues it causes. Lag is very unacceptable in any software and we should avoid it at all costs.

buyBread:
doesn't really lag me that much and my laptop is fairly old

PR #481, opened 2018-06-06
Add LWSS's legit filenaming scheme

Kr4ken-9:
Used in Fuzion, this renames libcathook.so to a more inconspicuous name and copies it to /usr/lib64 before injecting. This results in injecting files that look like this: /usr/lib64/libabrt_gui.so. Using this will eliminate a possible way of kisak detecting cathook, since VAC is known to look for names instead of sigs when they get desperate.


BenCat07:
Backtrace ver?

PR #491, opened 2018-06-08
CircleCI: Merge when prompted

TotallyNotElite:


update
```diff
@@ -2,10 +2,6 @@
 sudo echo Elevated
```
nullifiedcat: Is the elevation still needed?

install-all
```diff
@@ -12,24 +14,6 @@ sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y && sudo apt update && sud
 
 git fetch;git pull origin;git submodule update --remote --recursive
```
nullifiedcat: Why did you put all of this in a single line?

update
TotallyNotElite: No, the update script does not need to be elevated anymore.

install-all
TotallyNotElite: Because line 15 are responsible for the same thing. Updating the repo and its dependencies. See line 9, same thing, installing base-dependencies. Same with line 21.

nullifiedcat: It will be better if you split them to separate lines. It improves readability.

TotallyNotElite:
This PR requires some instructions to be followed. https://imgur.com/a/SyyWDrN

install-all
BenCat07: fixed @nullifiedcat

.circleci/config.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+jobs:
+    build:
+      docker:
+        - image: registry.gitlab.com/nullworks/cathook-ci-docker/ubuntu-cat-dependencies:build
+      steps:
+        - checkout
+        - run: 
+            name: "Compiling cathook"
+            command: git submodule update --init --recursive; mkdir build; cd build; cmake ..; make -j$(grep -c '^processor' /proc/cpuinfo); cd ..
```
Marc3842h: I haven't seen this but just a heads up, CircleCI only gives you a limited of cores than the actual machine reports by looking at the cpuinfo.

Looking at the CircleCI page, Cathook has these much resources available: img - which is 2 cores. Now most of CircleCI's machiens are 32 core servers (and their Docker sandbox reports the actual CPU core count of 32 instead of the 2 that you are actually allowed to use).

I have also seen that @TotallyNotElite has changed the make core count to `8`, which is still wrong. If cathooks build process uses more than 2 cores, it'll receive a `Hangup` error.

I had a similiar problem back with Spookware and talked to a CircleCI engineer to fix this and he told me exactly that:
img

PR #505, opened 2018-06-14
UCCCCCP Type B only, no double spam and small changes

TotallyNotElite:
Cat_ucccccp_encrypt and cat_ucccccp_decrypt will still use type a but also decrypt both a and b.


src/hacks/KillSay.cpp
```diff
@@ -64,6 +66,11 @@ std::string ComposeKillSay(IGameEvent *event)
     if (g_IEngine->GetPlayerForUserID(kid) != g_IEngine->GetLocalPlayer())
         return "";
     std::string msg = source->at(rand() % source->size());
+    while (msg == lastmsg)
```
nullifiedcat: Will result in deadlock if there is only one message to choose from

PR #590, opened 2018-10-02
This should fix engine prediction.

Roguezilla:
Put stuff like aimbot between those 2 func and stuff like bhop before them.


Roguezilla:
nvm

PR #593, opened 2018-10-05
Fix WTS not working after map change

LightyPon:
Unless this has some other purpose I'm not aware of, it's useless code which only breaks WTS and requires hud_reloadscheme to fix.


BenCat07:
Thamks

PR #607, opened 2018-10-15
Updated flags on pacman to sync package databases before installing deps

crazyredd:
Some users, including myself, received 404 errors when running the install-all script on Arch. The problem for me was that I needed to sync the package databases before attempting to install lib32-glew.

This change won't affect anyone who installs with no issues already but will allow anyone with the problem to install properly without any manual intervention.


BenCat07:
no

PR #685, opened 2019-01-26
Merge l0fg with master

l0fg:
More changes than expected
Overview:
- Added rage vote kick. Set RAGE state to player who attempted to kick friendly player. Attempt to vote kick players with RAGE state (controllable with `votelogger.autovote.no.rage`). Works only when `votelogger.autovote.no` is set to `true`
- Fix kick reason being ignored in vote kick command
- Add various party related signatures
- Use SendPartyChat function instead of constructing `say_party` command (that involved hacky `;` symbol replacements) in src/votelogger.cpp
- Fix `\` symbol not saved in settings
- Fix reporting mechanism reporting players with no steamID
- Added cat_pl_add_id. Now instead of using unreliable cat_pl_save/cat_pl_load you can maintain cfg file with all cat_pl_add_id (worthwhile note: files produced with cat_pl_save when ENABLE_VISUALS != 0 aren't compatible with ENABLE_VISUALS == 0 ones, because ENABLE_VISUALS != 0 introduces color field)
- Refactoring in some places. Mostly minor, but shouldTarget{,SteamId} is worth to double checking. shouldTarget{,SteamId} return true when it should target uses (i.e. reason equals to DO_NOT_IGNORE) and false otherwise

### DIFF
--- a/include/PlayerTools.hpp
+++ b/include/PlayerTools.hpp
@@ -27,8 +27,8 @@ enum class IgnoreReason
     OTHER
 };
 
-IgnoreReason shouldTargetSteamId(unsigned id);
-IgnoreReason shouldTarget(CachedEntity *player);
+bool shouldTargetSteamId(unsigned id);
+bool shouldTarget(CachedEntity *player);
 
 bool shouldAlwaysRenderEspSteamId(unsigned id);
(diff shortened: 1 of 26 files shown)


BenCat07:
nice, merging

PR #686, opened 2019-01-26
Improve drawing performance by using isurface instead of glez

TotallyNotElite:
Add new functions to drawing.cpp and make everything use it.


BenCat07:
no

TotallyNotElite:
> no

Excuse me wtf

PR #688, opened 2019-02-08
game glow cathook edition

Sx50:
flat glow instead of soft /shrug looks nicer
also fixed self rainbow glow not working and powerups not being toggleable 
image


BenCat07:
where are the blur scale go to

Sx50:
dead

PR #696, opened 2019-03-05
Fix Custom Item Schema Loading

hkva:
This patch re-adds the ability to load a custom item schema from `/opt/cathook/data/items_game.txt` using the command `cat_schema`


TotallyNotElite:
Thanks for your contribution!

BenCat07:
Yessssssss

BenCat07:
Thanks for this

PR #737, opened 2019-04-09
CMake Compiler Options

ghost-101:
_GCC Version: 8.2.1_

#### Each level includes the set from the previous
**LVL1 =** -O2
**LVL2 =** -O3 with Graphite and -fno-semantic-interposition
**LVL3 =** -Ofast and other non-compliant flags
**LTO1=** No Graphite
**LTO2 =** Graphite
**LTO3 =** Non-compliant flags

#### Grab available instruction sets:
``gcc -march=native -E -v - </dev/null 2>&1 | grep cc1``

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Optimize-Options - documents most in the CMake

#### Causes crashing upon injection, or after loading a map with LTO enabled
**From LVL3:** ``-funroll-loops``
**From LVL3:** ``-fvpt`` when combined with ``-fprofile-arcs`` and ``-fbranch-probabilities`` to effectively work

#### Causes crashing upon injection, or after loading a map:
**From -O2:** ``-fipa-cp -fipa-sra``

#### Causes features to fail, such as backtracking:
**From -O2:** ``-fschedule-insns2``

## Other notes
``-flto`` (Link Time Optimizations/LTO) can take in additional arguments for the count of processor threads to be utilized, as in ``-flto=8`` for 8 threads

Cheat functions need to be benchmarked with these flag combinations to see which are worth keeping around.

LVL3 and LTO3 are not a good idea to mainline due to violating ISO C++ compliance, and might not even bring tangible benefits.


BenCat07:
Uhm merge or not? What do you want 😕

ghost-101:
> Uhm merge or not? What do you want 😕
I'm making sure the flags set work out of the box, they don't right now.

ghost-101:
Looks okay to merge now, anything other than Debug and Release will always be experimental still, unless one of them becomes Release.

TotallyNotElite:
What exactly is the point of all of these compile options? Would be great if you could give us some info. Thanks. (I mean what does LTO3 mean for example)

PR #767, opened 2019-06-22
Request sudo at 0 

SylveonBottle:
image

### DIFF
--- a/update
+++ b/update
@@ -1,9 +1,7 @@
 #!/usr/bin/env bash
 
-sudo echo "got me some sudo?" > /dev/null
-
 if [ $SUDO_USER ]; then
     sudo -u $SUDO_USER ./scripts/updater
 else
-    ./scripts/updater
+    exec sudo /bin/bash "$0" "$@"
 fi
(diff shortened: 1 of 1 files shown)


update
```diff
@@ -1,9 +1,7 @@
 #!/usr/bin/env bash
 
-sudo echo "got me some sudo?" > /dev/null
-
 if [ $SUDO_USER ]; then
     sudo -u $SUDO_USER ./scripts/updater
 else
-    ./scripts/updater
+    exec sudo /bin/bash "$0" "$@"
```
TotallyNotElite: In my opinion this is a better way, since it should avoid the sudo prompt even after multiple minutes have passed. However, I'm not sure if we should force sudo rights for ./update. I will keep this pull request open in the meantime.

(NOTE: We should scrap /opt/cathook/data entirely and use a local directory instead)

TotallyNotElite:
Invalidated because 9f0a5d141012dc83af1fe0fe360d5d20e2d3fb94 completely removes the sudo requirement.

PR #808, opened 2019-08-31
visuals stuff

BenCat07:


src/core/logging.cpp
```diff
@@ -14,6 +14,7 @@
 #include "common.hpp"
 #include "hack.hpp"
 #include "MiscTemporary.hpp"
+#include "thread"
```
TotallyNotElite: Thread include is not required.

src/core/logging.cpp
```diff
@@ -34,8 +35,9 @@ void logging::Info(const char *fmt, ...)
 #if ENABLE_LOGGING
     if (shut_down)
         return;
-    logging::Initialize();
-
+    if (!handle.is_open())
+        logging::Initialize();
+    handle.flush();
```
TotallyNotElite: Why are you flushing before you actually write the message to the ofstream?

TotallyNotElite: Line 63 needs to be replaced with handle.flush() instead of handle.shutdown()

src/hack.cpp
```diff
@@ -375,6 +366,7 @@ static InitRoutineEarly nullify_textmode([]() {
     patch1.Patch();
     patch2.Patch();
 });
+static InitRoutine nullifiy_textmode2([]() { ReduceRamUsage(); });
```
TotallyNotElite: This second call to ReduceRamUsage is freezing preload catbots.

src/hooks/nographics.cpp
```diff
@@ -6,6 +6,7 @@
  */
 
 #include "common.hpp"
+#include "MiscTemporary.hpp"
```
TotallyNotElite: The RVar can be placed in this file instead of using an extern in MiscTemporary.hpp

src/hooks/nographics.cpp
```diff
@@ -72,3 +73,213 @@ static InitRoutine init_nographics([]() {
...
+static bool FSHook_Precache(const char *pFileName, const char *pPathID)
+{
+    return true;
+}
+
+static CatCommand debug_invalidate("invalidate_mdl_cache", "Invalidates MDL cache", []() { g_IBaseClient->InvalidateMdlCache(); });
+
+static hooks::VMTHook fs_hook{}, fs_hook2{};
+static bool hooked_fs = false;
+static void ReduceRamUsage()
```
TotallyNotElite: Function still causes freezes. 
image

src/hooks/nographics.cpp
BenCat07: at this point, yes it can

src/hooks/nographics.cpp
TotallyNotElite: Seems to work fine on my machine tho.

TotallyNotElite: Resolved.

BenCat07: Works, we just forgot some changes to compile options on catbots

src/hack.cpp
BenCat07: fixed

src/core/logging.cpp
BenCat07: fixed

src/core/logging.cpp
BenCat07: fixed

src/hooks/nographics.cpp
BenCat07: done.

PR #822, opened 2019-09-28
Fixes to crashes

BenCat07:
- BytePatch which would wrap over two or more pages no longer causes a crash
- The Scoreboard icons no longer crash the game (hotfix undone)


BenCat07:
The Former also has to be put onto backports

PR #823, opened 2019-09-28
Don't mark yourself as party state and autovote next map

BenCat07:


data/menu/nullifiedcat/misc/catbot.xml
```diff
@@ -55,6 +55,7 @@
             <AutoVariable width="fill" target="cat-bot.anti-autobalance" label="Anti Auto balance"/>
             <AutoVariable width="fill" target="misc.autodisguise" label="Autodisguise"/>
             <AutoVariable width="fill" target="cat-bot.anti-motd" label="Anti MOTD"/>
+            <AutoVariable width="fill" target="cat-bot.autovote-map" label="Autovote next map"/>
```
TotallyNotElite: "Autovote next map" does not do a very good job of explaining what this does. Maybe rename it to "Autovote for current map".

BenCat07: done

PR #838, opened 2019-10-19
Script updates

TotallyNotElite:
- Added migration to fix a CoTiRe error
- Add code to sync submodules if in non developer mode
- Don't run migrations in developer mode
- Update the developer script


TotallyNotElite:
Force push to rebase

PR #839, opened 2019-10-19
Add Anti party spam crash

BenCat07:


src/hooks/FireEventClientSide.cpp
```diff
@@ -12,8 +12,53 @@ extern void fire_event(IGameEvent *event);
...
+                // You're free, now flee
+                if (spam_timer.check(15000))
+                    messages_since_reset = 0;
+                // How about we don't send that anywhere m'lad
+                else
+                    return true;
+            }
+
+            // Not banned yet
+            if (messages_since_reset < 12)
```
TotallyNotElite: Use an else statement? Don't hardcode 12 here?

src/hooks/FireEventClientSide.cpp
```diff
@@ -12,8 +12,53 @@ extern void fire_event(IGameEvent *event);
...
+
+int messages_since_reset{ 0 };
+Timer spam_timer{};
+
 DEFINE_HOOKED_METHOD(FireEventClientSide, bool, IGameEventManager2 *this_, IGameEvent *event)
 {
+    if (enable_antispam)
+        if (!strcmp(event->GetName(), "party_chat"))
+        {
+            // You have violated the law, please sit in timeout for your sins
```
TotallyNotElite: Maybe consider making the comments less stupid.

src/hooks/FireEventClientSide.cpp
```diff
@@ -12,8 +12,53 @@ extern void fire_event(IGameEvent *event);
...
+            if (messages_since_reset >= *spam_rate)
+            {
+                // Way to reuse same timer
+                if (messages_since_reset == *spam_rate)
+                {
+                    spam_timer.test_and_set(0);
```
TotallyNotElite: Why are you using test_and_set when you don't use the output? Why not use .update()?

src/hooks/FireEventClientSide.cpp
```diff
@@ -12,8 +12,53 @@ extern void fire_event(IGameEvent *event);
...
+                    return true;
+            }
+
+            // Not banned yet
+            if (messages_since_reset < 12)
+            {
+                // Way to reset timer every 5 seconds
+                if (spam_timer.check(5000))
+                {
+                    spam_timer.test_and_set(0);
```
TotallyNotElite: What are you doing with this poor timer class? Why are you using `.check` AND `.test_and_set`? You could simply use `.test_and_set` and use its bool output.

src/hooks/FireEventClientSide.cpp
```diff
@@ -12,8 +12,53 @@ extern void fire_event(IGameEvent *event);
...
 DEFINE_HOOKED_METHOD(FireEventClientSide, bool, IGameEventManager2 *this_, IGameEvent *event)
 {
+    if (enable_antispam)
+        if (!strcmp(event->GetName(), "party_chat"))
+        {
+            // You have violated the law, please sit in timeout for your sins
+            if (messages_since_reset >= *spam_rate)
+            {
+                // Way to reuse same timer
+                if (messages_since_reset == *spam_rate)
```
TotallyNotElite: Consider using a bool value `spam_banned` instead of using hacks like this. I mean that code works, but it's a hack.

src/hooks/FireEventClientSide.cpp
BenCat07: My brain is running on air rn, sorry for that

src/hooks/FireEventClientSide.cpp
BenCat07: brain machine broke

src/hooks/FireEventClientSide.cpp
BenCat07: it just works

well yeah i could also add another timer to make it even less hacky

src/hooks/FireEventClientSide.cpp
BenCat07: i refuse (but ok i will)

BenCat07:
LGTM

PR #844, opened 2019-10-21
Add Load from string for settings manager + misc changes

BenCat07:
- Add a Missing Spellbook (which is actually an orb) to spellbook list

- Forward declare the queue_time Timer to allow Modules to stop catbots from exiting after 20 minutes of main menu idle

- Add a Way to load a config from a string (mostly for Modules)

- Add a bumper cart hack (thanks to UNKN0WN)

- Fix crashes and buggy text when drawing too much text (e.g. entity esp)

### DIFF
--- a/src/settings/SettingsIO.cpp
+++ b/src/settings/SettingsIO.cpp
@@ -132,6 +132,33 @@ bool settings::SettingsReader::loadFrom(std::string path)
     return true;
 }
 
+bool settings::SettingsReader::loadFromString(std::string stream)
+{
+    settings::SettingsReader loader{ settings::Manager::instance() };
+    if (stream == "")
+    {
+        logging::Info("cat_load: Can't access file!");
+        g_ICvar->ConsoleColorPrintf(MENU_COLOR, "CAT: cat_load: File doesn't exist / can't open file!\n");
+        return false;
+    }
+
+    for (auto &v : settings::Manager::instance().registered)
+    {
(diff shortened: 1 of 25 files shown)


src/settings/SettingsIO.cpp
```diff
@@ -132,6 +132,33 @@ bool settings::SettingsReader::loadFrom(std::string path)
     return true;
 }
 
+bool settings::SettingsReader::loadFromString(std::string stream)
+{
+    settings::SettingsReader loader{ settings::Manager::instance() };
+    if (stream == "")
+    {
+        logging::Info("cat_load: Can't access file!");
```
TotallyNotElite: What file?

src/settings/SettingsIO.cpp
```diff
@@ -132,6 +132,33 @@ bool settings::SettingsReader::loadFrom(std::string path)
...
+        g_ICvar->ConsoleColorPrintf(MENU_COLOR, "CAT: cat_load: File doesn't exist / can't open file!\n");
```
TotallyNotElite: Not a file

src/settings/SettingsIO.cpp
BenCat07: resolved

src/settings/SettingsIO.cpp
BenCat07: resolved

PR #847, opened 2019-10-22
Fix Binding issues on Keys

BenCat07:


TotallyNotElite:
What issue does this fix? Please be more specific.

BenCat07:
> What issue does this fix? Please be more specific.

I recently migrated to using the key format given out by sdl instead of tf2's, which comes with the advantage of preload not breaking keys. However i accidentally forgot the line for normal non mouse buttons, so it didn't work in the menu.

TotallyNotElite:
I still do not understand what the purpose of this PR is. What issue is it fixing? What didn't work in the menu?

robootgit:
Before, you couldn't bind any keys besides mouse buttons in the menu

PR #849, opened 2019-10-22
Fix #847 not working as intended

BenCat07:
i'm rarted


BenCat07:
one line install script vs one line commit

PR #855, opened 2019-10-25
Added a few color options I forgot about + slight menu changes

buyBread:
The UI Colors tab with the added options
  
Screenshot from 2019-10-25 20-17-23
  
Also changed Fake Lag and Spin Speed to be a slider along with moving some things around in the HvH tab for simplicity's sake.  
* Spin Speed min/max is -45 to 45 as above/below is quite literally just Random which is already an option.
* Fake Lag min/max is 0 to 14. Above 14 causes actual lag.
   
Screenshot from 2019-10-25 20-40-40
  
As well as making FOV Opacity and Slow Aimbot into sliders and moving some things to be more consistent.
* Slow Aimbot min/max is 0 to 30. More than 30 is just memeing and doesn't help you aim better at all, though I'd say 30 is already going into meme territory due to how slow it is.
  
Screenshot from 2019-10-25 20-40-20

### DIFF
--- a/data/menu/nullifiedcat/aimbot.xml
+++ b/data/menu/nullifiedcat/aimbot.xml
@@ -2,6 +2,7 @@
     <Box padding="12 6 6 6" width="content" height="content" name="Aimbot">
         <List width="150">
             <AutoVariable width="fill" target="aimbot.enable" label="Enable Aimbot"/>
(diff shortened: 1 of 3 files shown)


BenCat07:
15 still works just fyi.

buyBread:
Was a bit laggy for me

BenCat07:
> Was a bit laggy for me

Only if you yourself also have actual packet loss usually

PR #861, opened 2019-10-27
Update scripts

TotallyNotElite:
- Install git independently of other packages (Stable may have other dependencies than master. Also allows us to remove some package install hacks from nullworks/One-in-all-cathook-install)
- Improve the update scripts capabilities to recover a bad repo

### DIFF
--- a/install-all
+++ b/install-all
@@ -5,23 +5,6 @@ RUNCMD="sudo -u $RUNUSER"
 
 set -e
 
-# install all dependencies
-./scripts/dependencycheck
-
-#
-# Update cathook
-#
-if [ ! -d "./.git" ]; then
-    $RUNCMD git init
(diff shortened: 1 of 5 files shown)


install-all
```diff
@@ -5,23 +5,6 @@ RUNCMD="sudo -u $RUNUSER"
 
 set -e
 
-# install all dependencies
```
BenCat07: *confused screaming*

TotallyNotElite: ?

BenCat07: why is this gone entirely

TotallyNotElite: https://github.com/nullworks/cathook/pull/861/files#diff-5d98e6a5f9ed1ef89f93d94567219d9cR34
https://github.com/nullworks/cathook/pull/861/files#diff-f72b85a6905939b077ec7953f9594ad7R132

PR #865, opened 2019-11-01
viewmodel aimbot (skinny boi)

EricSmith17:
yeaaaaaaaaaaa

### DIFF
--- a/src/hooks/visual/OverrideView.cpp
+++ b/src/hooks/visual/OverrideView.cpp
@@ -11,6 +11,8 @@ static settings::Float override_fov_zoomed{ "visual.fov-zoomed", "0" };
 static settings::Float override_fov{ "visual.fov", "0" };
 static settings::Float freecam_speed{ "visual.freecam-speed", "800.0f" };
 static settings::Button freecam{ "visual.freecam-button", "<none>" };
+static settings : Boolean vm_aimbot{ "visual.vm-aimbot", "false" };
+
 bool freecam_is_toggled{ false };
 
 namespace hooked_methods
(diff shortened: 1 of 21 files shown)


src/hooks/visual/OverrideView.cpp
```diff
@@ -32,6 +34,61 @@ DEFINE_HOOKED_METHOD(OverrideView, void, void *this_, CViewSetup *setup)
...
+        } // reset
+
+        auto Viewmodel = (IClientEntity *) g_IEntityList->GetClientEntityFromHandle(CE_INT(LOCAL_E, netvar.hViewModel));
+
+        if (!Viewmodel)
+        {
+            return;
+        }
+
+        if (first || oAngle != oViewmodelAimbotAngle)
```
robootgit: "first" is ALWAYS going to be true because you set it to true on line 41

src/hooks/visual/OverrideView.cpp
```diff
@@ -32,6 +34,61 @@ DEFINE_HOOKED_METHOD(OverrideView, void, void *this_, CViewSetup *setup)
...
+
+        fClampAngle(deltaAngle);
+
+        Viewmodel->SetAbsAngles(setup->angles + deltaAngle);
+
+        timeremaining -= g_GlobalVars->frametime;
+
+        ViewmodelAimbotAngle = AngleLerp(setup->angles, oAngle, timeremaining / maxtime);
+
+        fClampAngle(ViewmodelAimbotAngle);
```
robootgit: You are calling flClampAngle and AngleLerp AFTER SetAbsAngles

src/hooks/visual/OverrideView.cpp
```diff
@@ -32,6 +34,61 @@ DEFINE_HOOKED_METHOD(OverrideView, void, void *this_, CViewSetup *setup)
...
+            oAngle        = ViewmodelAimbotAngle;
+            first         = false;
+            timeremaining = maxtime;
+        }
+
+        auto deltaAngle = ViewmodelAimbotAngle - setup->angles;
+
+        fClampAngle(deltaAngle);
+
+        Viewmodel->SetAbsAngles(setup->angles + deltaAngle);
```
robootgit: The method "SetAbsAngles" does not exist.

BenCat07:
i atleast fixed the compile issues, a crash, and the return statement issues, but this whole thing still doesn't work, you will have to look into why and fix it.

src/hooks/visual/OverrideView.cpp
EricSmith17: yes and ?

src/hooks/visual/OverrideView.cpp
EricSmith17: :face_palm: you

src/hooks/visual/OverrideView.cpp
EricSmith17: someoine said its defined in source SDK so i didnt knew that

EricSmith17:
thanks 👍

src/hooks/visual/OverrideView.cpp
robootgit: You are checking to see if it is true here. Why are you checking to see if it is true if it will always be true?

PR #876, opened 2019-11-07
Add ragdoll style overrides

robootgit:
This patch adds an option to change the way ragdolls look.

Video demo: https://youtu.be/9Ds83esfbKc

There is also an option to only modify ragdolls that are spawned from players that were killed by the local player. This is enabled by default.

Currently, it will only change the ragdoll style if it is not already set. If a ragdoll is already set to be a gib and you have the override set to a gold ragdoll, it will spawn as a gib. If people want, you could easily add an option to "force" the selected ragdoll style.

I tested this for about half an hour in casual and I found that it will _occasionally_ fail if the person you just killed was airborn. I don't know if this is an issue with the way I detect if the ragdoll belongs to a person the player just killed, or if it's a bug with the game.

### DIFF
--- a/src/hacks/Ragdolls.cpp
+++ b/src/hacks/Ragdolls.cpp
@@ -103,7 +103,7 @@ bool ragdollKillByLocal(void *ragdoll)
 void overrideGib(const CRecvProxyData *data, void *structure, void *out)
 {
     auto gib = reinterpret_cast<bool *>(out);
-    if (static_cast<int>(mode) == RagdollOverride_t::GIB && (!static_cast<bool>(only_local) || ragdollKillByLocal(structure)))
+    if (*mode == RagdollOverride_t::GIB && (!*only_local || ragdollKillByLocal(structure)))
         *gib = true;
     else
         *gib = data->m_Value.m_Int;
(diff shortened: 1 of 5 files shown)


robootgit:
...I have no idea why my commit shows up as @pongsarunb

BenCat07:
> ...I have no idea why my commit shows up as @pongsarunb

you have encountered fungwah bug

src/hacks/Ragdolls.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+static settings::Int mode{ "visual.ragdoll-mode", "0" };
+static settings::Boolean only_local{ "visual.ragdoll-only-local", "1" };
+
+/**
+ * Simple helper class for swapping out a RecvVarProxyFn
+ * and restoring it later on.
+ * You MUST call init(...) before calling setHook() or restore()
+ */
+class ProxyFnHook
```
BenCat07: Oh niceeeeeeee

src/hacks/Ragdolls.cpp
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+
+private:
+    bool hooked;
+    RecvPropRedef *prop;
+    RecvVarProxyFn original_fn;
+};
+
+// Ragdoll override style
```
BenCat07: Is there like a cowmangler one? (plasma + floats upwards)

src/hacks/Ragdolls.cpp
BenCat07: You could check if (proxy) in setHook and restore though to avoid issues in Shutdown possibly

src/hacks/Ragdolls.cpp
robootgit: There's no single variable for controlling the cow mangler effect. Maybe it's a combination of the existing variables? I can check.

Issue #862, opened 2019-10-28
[Suggestion] Add option to Auto-Reflect  to ignore specific projectiles


raphael-costa:
I don't know if this is possible but reflecting flares doesn't achieve much (wasted ammo most of the time) and it also looks a bit fishy. It would be interesting if we had an option to ignore it.

Stickies and Direct Hit rockets are also very hard to reflect.

TotallyNotElite:
+1

infoman-cor:
@TotallyNotElite added #901

PR #901, opened 2020-02-03
AutoReflect support for toggling several different projectile types

Rtas-17:
Mostly just modified IsEntStickyBomb and places where it's referenced to fit other projectiles. Feel like a lot of this could probably be made way more efficient, but it works I suppose

Created in response to #862

### DIFF
--- a/src/hacks/AutoReflect.cpp
+++ b/src/hacks/AutoReflect.cpp
@@ -16,25 +16,24 @@ static settings::Boolean legit{ "autoreflect.legit", "false" };
 static settings::Boolean dodgeball{ "autoreflect.dodgeball", "false" };
 static settings::Button blastkey{ "autoreflect.button", "<null>" };
 
+static settings::Boolean proj_default{ "autoreflect.default", "false" };
 static settings::Boolean stickies{ "autoreflect.stickies", "true" };
 static settings::Boolean pipes{ "autoreflect.pipes", "true" };
-static settings::Boolean flares{ "autoreflect.flares", "false"};
-static settings::Boolean arrows{ "autoreflect.arrows", "false"};
-static settings::Boolean jars{ "autoreflect.jars", "false"};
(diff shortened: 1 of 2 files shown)


src/hacks/AutoReflect.cpp
```diff
@@ -39,6 +51,88 @@ bool IsEntStickyBomb(CachedEntity *ent)
...
+{
+    if (ent->m_iClassID() == CL_CLASS(CTFProjectile_BallOfFire) || ent->m_iClassID() == CL_CLASS(CTFProjectile_EnergyBall))
+        return true;
+    else
+        return false;
+}
+
+bool IsEntCleaver(CachedEntity *ent)
+{
+    if (ent->m_iClassID() == CL_CLASS(CTFProjectile_BallOfFire) || ent->m_iClassID() == CL_CLASS(CTFProjectile_Cleaver))
```
BenCat07: Wait what why is the Dragons fury fireball a cleaver?

TotallyNotElite:
@Rtas-17 what's the status of this pull request?

src/hacks/AutoReflect.cpp
Rtas-17: Fixed that, my bad

Rtas-17:
Removed the "balls" option entirely, as no pointer seems to exist to the any of the CTFProjectile_Ball variants. Fixing something like that is far beyond my skill level, so here's a band-aid for the time being

TotallyNotElite:
image
this floor is made out of floor

Rtas-17:
Think I've done as asked, and have also set big projectiles to be reflected as default, as it can get quite tedious going through and enabling those (and there's a very high likelihood that if someone is enabling auto-reflect to begin with, they're gonna want at-least the big projectiles to be blasted).

TotallyNotElite:
image

TotallyNotElite:
Tested this a bit, should be fine.

PR #905, opened 2020-02-10
Improve Navbot

BenCat07:
WARNING: this bases on #902, so it needs to be merged first.

*Make sniper bots ignore DONT_HIDE flags
*Make bots path in a way to get from a to b quicker and avoid running into walls


BenCat07:
dead

PR #908, opened 2020-02-12
Allow custom sudo command via env vars

TotallyNotElite:
Preparation for Cathook GUI, since the user can't type in the password.


TotallyNotElite:
Thanks!

PR #913, opened 2020-02-13
Fixed Menu Capitalization

LightPower392:
It always bothered me that much of the capitalization throughout the main window was wrong. I went through all the tabs and fixed it.


LightPower392:
Given that for most of the strings the words after the first were capitalized, I assumed it was using title capitalization rules. See rule 6 of the url you linked (also the capitalization of the rules themselves).

PR #915, opened 2020-02-17
Fix movement.xml and add disguised option to chams

LightPower392:
The capitalization of movement.xml and autoheal.xml have been fixed. Also, there is an option for ignore disguised in chams

### DIFF
--- a/src/visual/EffectChams.cpp
+++ b/src/visual/EffectChams.cpp
@@ -16,6 +16,7 @@ namespace effect_chams
 static settings::Boolean flat{ "chams.flat", "false" };
 static settings::Boolean health{ "chams.health", "false" };
 static settings::Boolean teammates{ "chams.show.teammates", "false" };
+static settings::Boolean disguised{ "chams.show.disguised", "true" };
 static settings::Boolean players{ "chams.show.players", "true" };
 static settings::Boolean medkits{ "chams.show.medkits", "false" };
 static settings::Boolean ammobox{ "chams.show.ammoboxes", "false" };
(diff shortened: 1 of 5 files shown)


src/visual/EffectChams.cpp
```diff
@@ -24,7 +25,7 @@ static settings::Boolean stickies{ "chams.show.stickies", "true" };
 static settings::Boolean teammate_buildings{ "chams.show.teammate-buildings", "false" };
 static settings::Boolean recursive{ "chams.recursive", "true" };
 static settings::Boolean weapons_white{ "chams.white-weapons", "true" };
-static settings::Boolean legit{ "chams.legit", "false" };
+static settings::Boolean legit{ "chams.legit", "true" };
```
TotallyNotElite: Please avoid changing the defaults of variables without justification.

src/visual/EffectChams.cpp
```diff
@@ -233,6 +234,8 @@ bool EffectChams::ShouldRenderChams(IClientEntity *entity)
     case ENTITY_PLAYER:
         if (!players)
             return false;
+        if (!disguised && (IsPlayerDisguised(ent)))
```
TotallyNotElite: Should be expressed as `if (!disguised && IsPlayerDisguised(ent))` instead. The extra brackets are unnecessary.

LightPower392:
Removed the extra brackets from both chams and glow. Also reverted the default legit chams option.

TotallyNotElite:
Thanks for contributing! Your changes are currently on the master branch but not yet available to stable branch users. They will be included in the next stable update.

PR #916, opened 2020-02-17
Bhop chance and maximize sticky damage

LightPower392:
Two features that I thought were missing are a chance option to bhopping and a maximize sticky damage option. Bhop chance is the percent likelihood that a bhop succeeds. By default, auto sticky makes stickies blow up when someone is 130 hammer units away, which limited their damage in my play testing. This new option reduces the trigger radius to 80 hammer units, so they do more damage.


LightPower392:
I updated the min_dist to be on a slider from 0 to 130 with intervals of 5

PR #917, opened 2020-02-18
Radar Opacity and UI Tweaks

LightPower392:
This adds an options to control the opacity for the radar's background and also makes some UI changes such as giving pyro and spy one slot in Trigger with both of their functions.

Also I played with a catbot yesterday and it killed almost no one. Is this due to a coding issue or are the computers running them causing them to lag?


BenCat07:
The catbots ignore people that identify if you recall. Well someone made a tool for windows to identify with the catbots which makes it accessible to anyone really.

TotallyNotElite:
> Also I played with a catbot yesterday and it killed almost no one. Is this due to a coding issue or are the computers running them causing them to lag?

You can load a catbot config yourself and see if you can spot some issues.

PR #927, opened 2020-03-05
Add support for cathook gui binary mode

TotallyNotElite:
- Allows user specific data directory
- Remove dependency "glew" when running in imgui mode

This allows someone to compile a binary that should work almost anywhere (18.04+) because of the steam runtime.


BenCat07:
Have you checked if all the visual modes still compile?

Issue #750, opened 2019-06-02
Weapon based aimbot?


1337miro:
I would like to have aimbot on certain weapons (e.g the Sniper Rifle) and not on others (e.g the SMG)

TotallyNotElite:
Todo

TotallyNotElite:
So here's what I'm planning for this feature request:

- Each weapon slot will have its own aimbot settings.
- This will be implemented by creating a new class responsible for loading and saving specific RVar groups. (You don't want to do it the way AimTux fuzion does it does it :P https://github.com/LWSS/Fuzion/blob/master/src/Hacks/aimbot.cpp#L870)
- I don't know the least stupid way to save it into config files yet.

PR #929, opened 2020-03-07
Tooltip Update

LightPower392:
I worked on the tooltip system and got a working version that does most of what I think it should. If you hover over a menu element, it will print the tool tip if one is present in the xml, otherwise it will display the label.
One problem is that the background color is broken. The background is only a dot for me, and without it seeing the text of the tooltip can be hard. I couldn't figure out how to fix this.
Also, I will start working on adding the tooltips in the xml files.


BenCat07:
Fails to compile

LightPower392:
Thanks for letting me know, It should work now

BenCat07:
Thank you very much for this commit, i'd noticed tooltips were not working but not 100% sure how i would've gone about adding them (Especially because i did not have the time)

Reviewing now

BenCat07:
Now that i am thinking about this, @LightPower392 Why are you not using the existing Tooltip object?

LightPower392:
I tried a lot of different things and this is the only thing that worked. Do you know how to get `bb.width` and `bb.height` to work so the tooltip can have a background?

BenCat07:
Looking at the code it should work just fine (bb.width/bb.height that is)

BenCat07:
Don't forget to change the width and height mode too

LightPower392:
I added the background to the tooltip text, so it can be read easier now. All that is left to do is the XML files.

LightPower392:
I added a few example tooltips for the aimbot tab

LightPower392:
I added some tooltips, changed some menu positions, and added a keybind for thirdperson toggle. This fork is ready to pull now; I will now work on the aimbot.

TotallyNotElite:
I'll check this later.

TotallyNotElite:
Can you please keep this pull request focused on tooltips and create separate pull requests for the other changes like third person, aimbot, etc?

LightPower392:
The thirdperson keybind is removed. All that is changed here now is tooltip.cpp and various .xml files

TotallyNotElite:
I'll look at this PR tomorrow, kinda wasting a lot of time moving to matrix.org.

LightPower392:
If this fork can get pulled, I can show some of the changes I have made to the aimbot system.

TotallyNotElite:
Pretty busy right now, don't have time to test. you can always create a new branch and make a seperate PR.

BenCat07:
Some of the aimbot tooltips could be improved with proper descriptions

BenCat07:
"Lock Target"
tooltip:
"Lock Target"

LightPower392:
The tooltip defaults to just the label, and will only be different if a tooltip is set in the xml file. I started with some of the less clear options, but this will probably be an effort to get them all done.

Also, can do you like the idea of splitting aimbot settings into weapon slots?
Screenshot_2020-03-22_15-03-25

TotallyNotElite:
> The tooltip defaults to just the label

The default should be no tooltip.

> Also, can do you like the idea of splitting aimbot settings into weapon slots?

Check https://github.com/nullworks/cathook/issues/750, the main issue about this is storage of the settings, and I really don't want someone to write a half-baked solution for that. The real solution will probably be to switch to a JSON boost property tree and use that to store settings based on weapon id.

Also, do you ignore my advice on purpose? You don't need to make a new freaking GitHub account just to make a new pull request.

LightPower392:
Thanks for the suggestion, Elite. I changed it so that if a menu item has no defined tooltip in the XML, there simply won't be a tooltip. 
Also, are you saying that this stack overflow is wrong?
[https://stackoverflow.com/questions/28119733/how-can-i-make-a-second-fork-of-a-github-project](url)

LightPower392:
Thanks for the suggestion, Elite. The default is now no tooltip, and one will only be visible is set in the XML.
Also, is the solution for two pull requests from one account to create a new pull request from a second branch?

TotallyNotElite:
> Also, is the solution for two pull requests from one account to create a new pull request from a second branch?

Greetings! Yes indeed, you can create a second branch and use said branch to create a new pull request. It is generally recommended to create multiple pull requests for different features since it prevents working features from getting held up by a problem in a different feature.

LightPower392:
Does this look ready to approve, or are there changes you would like to be done?

TotallyNotElite:
@BenCat07 requesting review

include/visual/menu/menu/object/container/LabeledObject.hpp
```diff
@@ -27,10 +27,12 @@ class LabeledObject : public Container
 
     void setObject(std::unique_ptr<BaseMenuObject> &&object);
 
-    void setLabel(std::string text);
+    void setWithoutTooltip(std::string text1);
```
TotallyNotElite, line 30: These two functions should be "overloaded", meaning that they have the same name, but different parameters. I suggest that you keep the name "setLabel" and simply overload it, one with and one without setting the tooltip.

src/visual/menu/menu/Text.cpp
```diff
@@ -15,6 +15,21 @@ void zerokernel::Text::render()
...
+void zerokernel::Text::renderForTooltip()
+{
+    renderBackground(*zerokernel_tooltip::color_background);
+    renderBorder(*zerokernel_tooltip::color_border);
+#if ENABLE_IMGUI_DRAWING // needed, sorry
+    draw::String(bb.getContentBox().left() + text_x, bb.getContentBox().top() + text_y - 1, *color_text, data.c_str(), resource::font::base);
+#else
+    draw::String(bb.getContentBox().left() + text_x, bb.getContentBox().top() + text_y + 1, *color_text, data.c_str(), resource::font::base);
+#endif
+
```
TotallyNotElite, line 32: This is duplicate code, you could call zerokernel::Text::render instead.

src/visual/menu/menu/objects/Tooltip.cpp
```diff
@@ -28,10 +28,7 @@ void Tooltip::render()
     x += 6;
     move(x, y);
 
-    renderBackground(*zerokernel_tooltip::color_background);
-    renderBorder(*zerokernel_tooltip::color_border);
-
-    text.render();
```
TotallyNotElite, line 34: Why was it necessary to move this out of Tooltip.cpp in the first place?

LightPower392:
I did the overloading of the functions, but the rendering of the tooltip's background has to stay that way. I don't know exactly why, but the `renderBackground` and `renderBorder` have to be in Text.cpp not Tooltip.cpp to work

TotallyNotElite:
this pull request just confuses me to be honest.

LightPower392:
Almost all of the files changed are .XML files that are changed to add in a tooltip description. The .cpp and .hpp files fix the tooltips for menu elements, and this is done to make the functionality of features clearer.

TotallyNotElite:
I'll look into it myself later and copy the xml files over. I've got some other bugs to fix right now tho.

TotallyNotElite:
Replaced by #995

PR #934, opened 2020-03-18
Thirdperson and Radar Options

LightPower1:
This is LightPower392. I think making a new account is the easiest way to create a second fork. This adds two options to thirdperson: it adds a keybind and allows you to auto disable it when zoomed. It also lets you use a circle radar. 
Also, I am working on adding the skinchanger to the UI if you want to see it.


TotallyNotElite:
Did you ignore my advice to create a new branch on purpose? Whatever.

BenCat07:
Tested and working

PR #945, opened 2020-04-05
Add Datacenter Selector

BenCat07:
EDIT: also fixed namestealer stealing friendly players names, and adds a way to split bots into chunks using an ipc exec


src/hacks/DataCenter.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+    regions = boost::join(new_regions, ",");
+}
+
+static void Init()
+{
+    enable.installChangeCallback([](settings::VariableBase<bool> &, bool on) { Hook(on); });
+    regions.installChangeCallback([](settings::VariableBase<std::string> &, std::string regions) { OnRegionsUpdate(regions); });
+    restrict.installChangeCallback([](settings::VariableBase<bool> &, bool) { Refresh(); });
+    auto create             = (void *(*) ()) e8call_direct(gSignatures.GetClientSignature("E8 ? ? ? ? 85 C0 89 ? ? ? ? ? 74 ? 8B ? ? ? ? ? 80 ? ? ? ? ? 00"));
```
TotallyNotElite, line 224: Please avoid searching for signatures on init in things that are default off. You can create a second init function that gets run if a bool is not set when it's enabled.

BenCat07: True. willdo

BenCat07: Addressed

PR #950, opened 2020-04-09
Config List in Menu

LightPower392:
This pull request adds a way to create, load, and delete configs through the menu. 
config
There are three text inputs where you enter the name of a config (no .conf). After pressing enter, the box automatically becomes empty. If the save or delete are used, then the list will update to reflect the change.


TotallyNotElite:
Same criticism as in the issue.

LightPower392:
The reason why there are so many commits for this pull request is because I branched the tooltip fork. Could you tell me what the alternative to an input field is?

TotallyNotElite:
Do you mind if I push to your branch, so I can fix this commit mess?

LightPower392:
Sure

TotallyNotElite:
There we go.

include/visual/menu/menu/special/ConfigsManagerList.hpp
```diff
@@ -0,0 +1,46 @@
...
+    void recursiveWork(TreeNode &node, size_t depth);
+
+    void construct();
+
+    void addVariable(std::string name, bool registered);
+
+    void removeVariables();
+
+
+    //
```
BenCat07, line 41: \\

src/settings/SettingCommands.cpp
```diff
@@ -34,25 +50,25 @@ static std::string getAutoSaveConfigPath()
     time_info = localtime(&current_time);
     strftime(timeString, sizeof(timeString), "%Y-%m-%d %H:%M", time_info);
 
-    DIR *config_directory = opendir(paths::getConfigPath().c_str());
+    DIR *config_directory = opendir(DATA_PATH "/configs");
```
BenCat07, line 53: This should continue using paths::getConfigPath()

src/visual/menu/GuiInterface.cpp
```diff
@@ -119,7 +120,7 @@ static void initPlayerlist()
 
 static void load()
 {
-    zerokernel::Menu::instance->loadFromFile(paths::getDataPath("/menu"), "menu.xml");
```
BenCat07, line 122: Same here for DataPath, and everywhere else paths:: was deleted

PR #954, opened 2020-04-11
Small bugfixes

BenCat07:


BenCat07:
debug.engine-pred-others was not meant to be used at all, pp.draw should be used instead

src/prediction.cpp
```diff
@@ -45,69 +43,12 @@ bool PerformProjectilePrediction(CachedEntity *target, int hitbox)
...
-            }
-        }
-}
-
 Vector Predict(Vector &pos, Vector &vel, Vector acceleration, std::optional<float> grounddistance, float &time)
 {
     PROF_SECTION(PredictNew);
     Vector result = pos;
-    // Quadratic + linear function to map position to time (v+a*t)*t
+    result += (acceleration / 2.0f) * pow(time, 2) + vel * time;
```
TotallyNotElite, line 50: https://youtu.be/-AQfQFcXac8?t=134

src/prediction.cpp
```diff
@@ -118,10 +59,9 @@ Vector PredictStep(Vector pos, Vector &vel, Vector acceleration, std::pair<Vecto
 {
     PROF_SECTION(PredictNew)
     Vector result = pos;
-    // Quadratic + linear function to go one step into the future
 
+    result += (acceleration / 2.0f) * pow(steplength, 2) + vel * steplength;
```
TotallyNotElite, line 63: https://youtu.be/-AQfQFcXac8?t=134

src/prediction.cpp
BenCat07: i mean sure i can also not use pow() if you want

PR #955, opened 2020-04-12
Fix and add tooltips

TotallyNotElite:
Tooltip fixes are mine, full credit for XML changes to @LightPower392.

image


TotallyNotElite:
I really do not like this capitalization in the tooltips.

TotallyNotElite:
I can't be bothered to find a proper fix for the string issue in IMGUI. I just increased the max string size so this doesn't happen anymore.

PR #956, opened 2020-04-14
T diff requests

LightPower392:
This pull request adds the two changed of @TDiff: it makes autostab not stab invulnerable targets and adds an option to remove taunts.
Also:
  It adds an option to hide team buildings on the radar. Currently even after disabling teammates, team buildings still show.
  It adds an option to show backtrack chams for 1/5 of the ticks. When I made it draw a cham for every tick, it had a huge fps loss. I tested and found that showing 20% of the ticks had little impact on fps but had the advantages of showing every tick.
  It adds two more chams options (in addition to normal and flat) that are less graphically intrusive.


TDiffff:
Hey, that was fast! Thanks :D

TotallyNotElite:
Can't accept this, this includes the config system changes. Please remove these commits.

LightPower392:
Does this look better?

TotallyNotElite:
Great, now just rebase onto master to fix the conflicts.

LightPower392:
done

PR #961, opened 2020-04-16
Add a Warp hack

BenCat07:
Seems i also accidentally formatted stuff but i'm leaving it in here

### DIFF
--- a/include/hooks/HookedMethods.hpp
+++ b/include/hooks/HookedMethods.hpp
@@ -10,7 +10,7 @@
 #include "SeedPrediction.hpp"
 #if ENABLE_VISUALS
 union SDL_Event;
-struct SDL_Window;
+class SDL_Window;
 #endif
 
 #define DECLARE_HOOKED_METHOD(name, rtype, ...) \
@@ -43,6 +43,7 @@ DECLARE_HOOKED_METHOD(FireGameEvent, void, void *, IGameEvent *);
 // IBaseClient
 DECLARE_HOOKED_METHOD(DispatchUserMessage, bool, void *, int, bf_read &);
 DECLARE_HOOKED_METHOD(IN_KeyEvent, int, void *, int, ButtonCode_t, const char *);
+DECLARE_HOOKED_METHOD(FrameStageNotify, void, void *, ClientFrameStage_t);
 // IInput
 DECLARE_HOOKED_METHOD(GetUserCmd, CUserCmd *, IInput *, int);
 // INetChannel
@@ -73,8 +74,6 @@ DECLARE_HOOKED_METHOD(ChatPrintf, void, CHudBaseChat *, int, int, const char *, 
 DECLARE_HOOKED_METHOD(OverrideView, void, void *, CViewSetup *);
 // IStudioRender
 DECLARE_HOOKED_METHOD(BeginFrame, void, IStudioRender *);
-// IBaseClient
-DECLARE_HOOKED_METHOD(FrameStageNotify, void, void *, ClientFrameStage_t);
 // SDL
 DECLARE_HOOKED_METHOD(SDL_GL_SwapWindow, void, SDL_Window *);
 DECLARE_HOOKED_METHOD(SDL_PollEvent, int, SDL_Event *);
--- a/include/hooks/HookedMethods.hpp
+++ b/include/hooks/HookedMethods.hpp
@@ -10,7 +10,7 @@
 #include "SeedPrediction.hpp"
 #if ENABLE_VISUALS
 union SDL_Event;
-class SDL_Window;
+struct SDL_Window;
 #endif
 
 #define DECLARE_HOOKED_METHOD(name, rtype, ...) \
--- a/src/hacks/Warp.cpp
+++ b/src/hacks/Warp.cpp
@@ -60,7 +60,7 @@ public:
 int GetWarpAmount()
 {
     static auto sv_max_dropped_packets_to_process = g_ICvar->FindVar("sv_max_dropped_packets_to_process");
-    return warp_override ? *warp_override : sv_max_dropped_packets_to_process->GetInt();
+    return warp_override ? warp_override : sv_max_dropped_packets_to_process->GetInt();
 }
 bool should_warp = true;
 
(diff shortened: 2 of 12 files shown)


LightPower392:
Interesting. Do you think this could be used to auto-dodge high speed non-explosive projectiles like arrows or sandman balls?

BenCat07:
That may be possible

include/hooks/HookedMethods.hpp
```diff
@@ -10,7 +10,7 @@
 #include "SeedPrediction.hpp"
 #if ENABLE_VISUALS
 union SDL_Event;
-struct SDL_Window;
+class SDL_Window;
```
TotallyNotElite, line 13: ... uh? ok??

src/hacks/Warp.cpp
```diff
@@ -0,0 +1,244 @@
...
+#include "offsets.hpp"
+#include "globals.h"
+#include "Warp.hpp"
+#include "HookTools.hpp"
+#include "usercmd.hpp"
+#include "localplayer.hpp"
+#include "netvars.hpp"
+#include "entitycache.hpp"
+#include "conditions.hpp"
+#include "velocity.hpp"
```
TotallyNotElite, line 21: Are you scared of including common.hpp?

src/hacks/Warp.cpp
```diff
@@ -0,0 +1,244 @@
...
+    bf_write m_DataOut;
+};
+
+int GetWarpAmount()
+{
+    static auto sv_max_dropped_packets_to_process = g_ICvar->FindVar("sv_max_dropped_packets_to_process");
+    if (!warp_override)
+        return sv_max_dropped_packets_to_process->GetInt();
+    else
+        return warp_override;
```
TotallyNotElite, line 66:
```suggestion
    return warp_override ? *warp_override : sv_max_dropped_packets_to_process->GetInt();
```

src/hacks/Warp.cpp
```diff
@@ -0,0 +1,244 @@
...
+        return;
+    warp_override = 0;
+    if (!warp_key.isKeyDown())
+    {
+        warp_last_tick = false;
+        Vector velocity{};
+        velocity::EstimateAbsVelocity(RAW_ENT(LOCAL_E), velocity);
+
+        // Bunch of checks, if they all pass we are standing still
+        if (velocity.IsZero() && !HasCondition<TFCond_Charging>(LOCAL_E) && !current_user_cmd->forwardmove && !current_user_cmd->sidemove && !current_user_cmd->upmove && (CE_INT(LOCAL_E, netvar.iFlags) & FL_ONGROUND) && !(current_user_cmd->buttons & (IN_ATTACK | IN_ATTACK2)))
```
TotallyNotElite, line 154: Aren't most of these checks redundant?

src/hacks/Warp.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // Warp peaking
+    else if (warp_peek)
+    {
+        // We Have warp
+        if (warp_amount)
+        {
+            // Warped last tick, time to reverse
+            if (warp_last_tick)
+            {
+                // Wait 1 tick before tping back
```
TotallyNotElite, line 180: "tping" should be replaced with warping

src/hacks/Warp.cpp
```diff
@@ -0,0 +1,244 @@
...
+            // Use everxy xth tick for charging
+            if (!(tickcount % *warp_movement_ratio))
+                should_charge = true;
+            move_last_tick = true;
+        }
+    }
+    // Warp peaking
+    else if (warp_peek)
+    {
+        // We Have warp
```
TotallyNotElite, line 174: I don't know of a language where the verb is capitalized.

include/hooks/HookedMethods.hpp
BenCat07: we had that wrong

src/hacks/Warp.cpp
BenCat07: no but you said i shouldn't unless needed

src/hacks/Warp.cpp
BenCat07: i mean sure

src/hacks/Warp.cpp
BenCat07: nope

src/hacks/Warp.cpp
BenCat07: fair enough

src/hacks/Warp.cpp
BenCat07: bruh

src/hacks/Warp.cpp
TotallyNotElite: When are most of these checks not redundant

include/hooks/HookedMethods.hpp
TotallyNotElite: Let me remind you that there is no difference between a struct and a class, except that a struct is public by default and a class is private by default.

src/hacks/Warp.cpp
BenCat07: - you are not currently moving and hit w
- you are a demoman who stood still and then starts charging
- obviously do not want to choke packets when you try to move
- don't want to choke all packets when in the air (would make you get stuck and stuff would glitch)
- don't want to prevent attacking

include/hooks/HookedMethods.hpp
BenCat07: well i saw a yellow warning up there basically saying "it's fine in gcc and clang but other compilers will rip your head off"

TotallyNotElite:
Oh, lol I thought that was an rvar.

BenCat07:
> Oh, lol I thought that was an rvar.

And i didn't read properly

include/hooks/HookedMethods.hpp
TotallyNotElite: Where did you see that?

BenCat07: qtcreator,
HookedMethods.hpp:13:1: warning: struct 'SDL_Window' was previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI
imrenderer.hpp:20:7: note: previous use is here

actually nevermind, only on ms stuff

BenCat07:
More changes to come

PR #962, opened 2020-04-16
Add more vaccinator autoshoot checks and ESP details

surepy:
just a minor thing that bugged me before

### DIFF
--- a/src/hacks/ESP.cpp
+++ b/src/hacks/ESP.cpp
@@ -1207,13 +1207,30 @@ void _FASTCALL ProcessEntity(CachedEntity *ent)
                     if (IsPlayerInvulnerable(ent))
                         AddEntityString(ent, "*INVULNERABLE*");
                     // Vaccinator
+                    // Vaccinator
                     if (HasCondition<TFCond_UberBulletResist>(ent))
                     {
-                        AddEntityString(ent, "*VACCINATOR*");
+                        AddEntityString(ent, "*BULLET VACCINATOR*", colors::FromRGBA8(220, 220, 220, 255));
                     }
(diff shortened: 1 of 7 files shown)


LightPower392:
I'd like to see more weapons included. Also, you can look up a weapon on the tf2 wiki and search for its damage type.

surepy:
> I'd like to see more weapons included. Also, you can look up a weapon on the tf2 wiki and search for its damage type.

added Pomson 6000; I don't know what else I am missing.

for The Loose Cannon (CTFCannon) I'm having trouble deciding if I should add or not.
because the initial knockback damage does not count as a blast, pretty sure.

LightPower392:
I meant adding projectiles like the direct hit and the grenade launcher.

surepy:
> I meant adding projectiles like the direct hit 

```
case CL_CLASS(CTFRocketLauncher_DirectHit):
```
> and the grenade launcher.

```
case CL_CLASS(CTFGrenadeLauncher):
```

it's there tho

src/hacks/ESP.cpp
```diff
@@ -1207,13 +1207,30 @@ void _FASTCALL ProcessEntity(CachedEntity *ent)
                     if (IsPlayerInvulnerable(ent))
                         AddEntityString(ent, "*INVULNERABLE*");
                     // Vaccinator
+                    // Vaccinator
```
BenCat07, line 1210: comment

TotallyNotElite:
Let me check something real quick.

surepy:
note to self: check if system time is always correct so time travel on commits don't happen

LightPower392:
lol. I think this looks good

PR #964, opened 2020-04-20
Config UI

LightPower392:
I closed the previous pull request, because it was getting too many commits. Also, I added a way to rename configs in the menu.
configui


TotallyNotElite:
Why was this closed?

PR #969, opened 2020-04-23
Fixed Bone ESP and More

LightPower392:
This Pull Request:
-fixes bone ESP and adds an option to color it according to player state
-adds an "assist only" option to aimbot that makes it only active if the mouse has moved in the last half second. This is useful for playing legit.
-adds a radar option for team buildings

Also, the bytepatch for keeping the sniper rifle breaks the command `zoom_sensitivity_ratio`

### DIFF
--- a/src/hacks/ESP.cpp
+++ b/src/hacks/ESP.cpp
@@ -171,9 +171,9 @@ struct bonelist_s
         }
     }
 
-    void _FASTCALL Draw(CachedEntity *enti, const rgba_t &color)
+    void _FASTCALL Draw(CachedEntity *ent, const rgba_t &color)
     {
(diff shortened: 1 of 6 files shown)


TotallyNotElite:
I can't approve this because it includes changes from other pull requests.

LightPower392:
done

src/hacks/ESP.cpp
```diff
@@ -171,9 +171,9 @@ struct bonelist_s
         }
     }
 
-    void Draw(CachedEntity *ent, const rgba_t &color)
+    void _FASTCALL Draw(CachedEntity *enti, const rgba_t &color)
```
BenCat07, line 174: Why renamed to enti? (Just nitpicking)

PR #972, opened 2020-04-23
Add NoSpread

BenCat07:
Does as it says. Only for projectiles though (Until we find a way to crack the server seed and or float time)


BenCat07:
Yeah it will and i will have to fix that

BenCat07:
> I think in nospread.cpp it should say april, not august lol

Well yes it should but let's pretend we didn't see that

PR #974, opened 2020-04-24
Nullnexus temp deadlock and off by one error fixes

TotallyNotElite:
This includes some high priority fixes that should be fast tracked to stable ASAP.

### DIFF
--- a/src/hacks/Exploits.cpp
+++ b/src/hacks/Exploits.cpp
@@ -148,7 +148,7 @@ void CM()
     case STEP_HEAVY:
         if (commands_timer.test_and_set(1000))
         {
-            g_IEngine->ClientCmd_Unrestricted("join_class heavyweapons");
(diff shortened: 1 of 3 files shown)


src/hacks/Exploits.cpp
```diff
@@ -148,7 +148,7 @@ void CM()
     case STEP_HEAVY:
         if (commands_timer.test_and_set(1000))
         {
-            g_IEngine->ClientCmd_Unrestricted("join_class heavyweapons");
+            g_IEngine->ClientCmd_Unrestricted("join_class heavyweapons;use *");
```
BenCat07, line 151: This too?

TotallyNotElite: Yes this too, improves reliability a lot somehow. I get it first try mostly now.

PR #977, opened 2020-04-28
Warp changes

BenCat07:
- Fix issue with charge in jump and active charge
- Add A demoknight mode

### DIFF
--- a/include/reclasses/CTFPlayerShared.hpp
+++ b/include/reclasses/CTFPlayerShared.hpp
@@ -30,6 +30,7 @@ public:
         // TODO signature
         return false;
     }
+    // Get max charge turning (yaw) to prevent glitchy movement
     inline static float CalculateChargeCap(CTFPlayerShared *self)
     {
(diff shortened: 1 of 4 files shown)


include/reclasses/CTFPlayerShared.hpp
```diff
@@ -25,5 +30,12 @@ class CTFPlayerShared
         // TODO signature
         return false;
     }
+    inline static float CalculateChargeCap(CTFPlayerShared *self)
```
TotallyNotElite, line 33:
```suggestion
    // Get max charge turning (yaw) to prevent glitchy movement
    inline static float CalculateChargeCap(CTFPlayerShared *self)
```

src/hacks/MiscAimbot.cpp
```diff
@@ -511,9 +512,45 @@ static void CreateMove()
...
+        if (fVal > flChargeYawCap)
+            return flChargeYawCap;
+        else if (fVal < -flChargeYawCap)
+            return -flChargeYawCap;
+    }
+
+    return fVal;
+}
+
+#define foffset(p, i) ((unsigned char *) &p)[i]
```
TotallyNotElite, line 537: We really should be using this more, huh?

BenCat07: i mean we should put this in helpers lol

PR #978, opened 2020-04-29
Bytepatch Fixes

BenCat07:
Make bullet traces only init when used, and make killfeed color be off by default, also rework bytepatch to do all the searching logic at patch time instead of construction.

### DIFF
--- a/src/hacks/KillfeedColor.cpp
+++ b/src/hacks/KillfeedColor.cpp
@@ -6,7 +6,7 @@
 #if !ENFORCE_STREAM_SAFETY
 namespace hacks::tf2::killfeed
 {
-static settings::Boolean enable{ "visual.killfeedcolor.enable", "false" };
+static settings::Boolean enable{ "visual.killfeedcolor.enable", "true" };
 typedef void (*GetTeamColor_t)(Color *, void *, int, int);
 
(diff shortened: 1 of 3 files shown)


src/hacks/KillfeedColor.cpp
```diff
@@ -6,7 +6,7 @@
 #if !ENFORCE_STREAM_SAFETY
 namespace hacks::tf2::killfeed
 {
-static settings::Boolean enable{ "visual.killfeedcolor.enable", "true" };
+static settings::Boolean enable{ "visual.killfeedcolor.enable", "false" };
```
TotallyNotElite, line 9: Huh?

BenCat07: Don't enable by default, idk why it was, if you want i can change it back to true

TotallyNotElite: idk i kinda like it that way

TotallyNotElite: And cause it doesn't impact gameplay, I'm fine with it being true by default

TotallyNotElite, line 9:
```suggestion
static settings::Boolean enable{ "visual.killfeedcolor.enable", "true" };
```

PR #983, opened 2020-05-02
Crash reporter: Send address of non cathook libs too

TotallyNotElite:


report-crash
```diff
@@ -96,7 +96,7 @@ append_crumb() {
 
 while read p; do
     lib="$(echo "$p" | cut -f 1)"
-    addr="$(echo $p | cut -f 2 -d " ")"
```
BenCat07, line 99: #quote(lives)matter

TotallyNotElite: Not our issue, closing

PR #997, opened 2020-05-08
Make naming more consistent and fix typo

micwoj92:


micwoj92:
How does it break it? Also what do you mean by "default killsay file" isnt it just killsays.txt? Sorry for dumb question

BenCat07:
> How does it break it? Also what do you mean by "default killsay file" isnt it just killsays.txt? Sorry for dumb question

You changed the default rvar values.

micwoj92:
Ah sorry,seemed fine toe, i shouldnt really touch it when i only know bash and basic python, sorry for making problem lol

PR #1000, opened 2020-05-09
Interface for "Ping Reducer"

aUniqueUser:
Interface for setting cl_cmdrate to -1 so I don't have to type it in console every time.

### DIFF
--- a/src/hacks/Misc.cpp
+++ b/src/hacks/Misc.cpp
@@ -56,6 +56,7 @@ static void tryPatchLocalPlayerShouldDraw(bool after)
 static Timer anti_afk_timer{};
 static int last_buttons{ 0 };
 
+static int oldCmdRate;
 static int currentCmdRate()
 {
     static ConVar *cl_cmdrate = g_ICvar->FindVar("cl_cmdrate");
@@ -340,18 +341,6 @@ void Draw()
                 AddSideString(format(info.name, " ", observermode));
             }
         }
-    }
-    if (ping_reducer && currentCmdRate() != -1)
-    {
-        ConVar *cmdrate    = g_ICvar->FindVar("cl_cmdrate");
-        cmdrate->m_fMaxVal = 999999999.9f;
-        cmdrate->m_fMinVal = -999999999.9f;
-        cmdrate->SetValue(-1);
-    }
-    if (!ping_reducer && currentCmdRate() != 66) // Hopefully everyone is using 66 its 2020 ffs
-    {
-        ConVar *cmdrate = g_ICvar->FindVar("cl_cmdrate");
-        cmdrate->SetValue(66);
     }
     if (!debug_info)
         return;
@@ -813,6 +802,24 @@ static InitRoutine init_pyrovision([]() {
             cart_patch1.Shutdown();
             cart_patch2.Shutdown();
         }
+    });
+    ping_reducer.installChangeCallback([](settings::VariableBase<bool> &, bool after) {
+        if (after && currentCmdRate() != -1)
+        {
+            ConVar *cmdrate = g_ICvar->FindVar("cl_cmdrate");
+            oldCmdRate = cmdrate->GetInt();
+
+            cmdrate->m_fMaxVal = 999999999.9f;
+            cmdrate->m_fMinVal = -999999999.9f;
+            cmdrate->SetValue(-1);
+        }
+        if (!after && currentCmdRate() != oldCmdRate)
+        {
+            ConVar *cmdrate = g_ICvar->FindVar("cl_cmdrate");
+            cmdrate->SetValue(oldCmdRate);
+        }
+        if (!oldCmdRate)
+            oldCmdRate = currentCmdRate();
     });
 #endif
 });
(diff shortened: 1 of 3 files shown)


micwoj92:
Wouldn't it break on 100 tick servers?

BenCat07:
looking at source

BenCat07:
fuck, like i was going to say, looking at source leak

BenCat07:
Please github stop placing the close button on my keyboard. Like i was saying, looking at the source leak it should reduce ping by 500 regardless of server

src/hacks/Misc.cpp
```diff
@@ -334,6 +341,18 @@ void Draw()
...
         }
     }
+    if (ping_reducer && currentCmdRate() != -1)
+    {
+        ConVar *cmdrate    = g_ICvar->FindVar("cl_cmdrate");
+        cmdrate->m_fMaxVal = 999999999.9f;
+        cmdrate->m_fMinVal = -999999999.9f;
+        cmdrate->SetValue(-1);
+    }
+    if (!ping_reducer && currentCmdRate() != 66) // Hopefully everyone is using 66 its 2020 ffs
```
BenCat07, line 351: How about you back up the Cmdrate in the above and compare it here? (if none is set, set it to the current one)

src/hacks/Misc.cpp
```diff
@@ -33,6 +33,9 @@ static settings::Boolean auto_balance_spam{ "misc.auto-balance-spam", "false" };
 static settings::Boolean nopush_enabled{ "misc.no-push", "false" };
 static settings::Boolean dont_hide_stealth_kills{ "misc.dont-hide-stealth-kills", "true" };
 static settings::Boolean unlimit_bumpercart_movement{ "misc.bumpercarthax.enable", "true" };
+static settings::Boolean ping_reducer{ "misc.ping-reducer.enable", "false" };
+
+settings::Int force_ping{ "misc.ping-reducer.target", "" };
```
TotallyNotElite, line 38: Set this to 0 explicitly by default

src/hacks/Misc.cpp
```diff
@@ -55,6 +58,14 @@ static void tryPatchLocalPlayerShouldDraw(bool after)
...
 static int last_buttons{ 0 };
 
+static int oldCmdRate;
+static int counter = 0;
+static int currentCmdRate()
+{
+    static ConVar *cl_cmdrate = g_ICvar->FindVar("cl_cmdrate");
+    return cl_cmdrate->GetInt();
+}
+
```
TotallyNotElite: Remove this

src/hacks/Misc.cpp
```diff
@@ -262,6 +273,29 @@ void CreateMove()
...
         else
             teammatesPushaway = g_ICvar->FindVar("tf_avoidteammates_pushaway");
+
+        // We only want this to run one time so we can get the right cmdrate.
+        if (counter == 0)
+        {
+            ConVar *cmdrate = g_ICvar->FindVar("cl_cmdrate");
+            oldCmdRate      = cmdrate->GetInt();
+            counter++;
+        }
```
TotallyNotElite: remove this, useless

src/hacks/Misc.cpp
```diff
@@ -262,6 +273,29 @@ void CreateMove()
...
+
+        // Custom forced ping
+        if (ping_reducer && *force_ping > 0)
+        {
+            ConVar *cmdrate = g_ICvar->FindVar("cl_cmdrate");
+            int ping        = g_pPlayerResource->GetPing(g_IEngine->GetLocalPlayer());
+            if (*force_ping <= ping)
```
TotallyNotElite, line 281: Only set this if cmdrate != -1 and then set oldcmdrate

src/hacks/Misc.cpp
```diff
@@ -795,6 +829,20 @@ static InitRoutine init_pyrovision([]() {
...
         }
     });
+    ping_reducer.installChangeCallback([](settings::VariableBase<bool> &, bool after) {
+        if (after && currentCmdRate() != -1)
+        {
+            ConVar *cmdrate    = g_ICvar->FindVar("cl_cmdrate");
+            cmdrate->m_fMaxVal = 999999999.9f;
+            cmdrate->m_fMinVal = -999999999.9f;
+            cmdrate->SetValue(-1);
+        }
```
TotallyNotElite, line 830: Remove this, useless

src/hacks/Misc.cpp
```diff
@@ -795,6 +829,20 @@ static InitRoutine init_pyrovision([]() {
...
+        if (!after && currentCmdRate() != oldCmdRate)
```
TotallyNotElite, line 831: You are already getting the convar anyway, why not just use the value of that instead of using currentCmdRate. You also need the handle the case where oldCmdRate may not be initialized (set) yet

src/hacks/Misc.cpp
```diff
@@ -262,6 +267,21 @@ void CreateMove()
...
+            ConVar *cmdrate = g_ICvar->FindVar("cl_cmdrate");
+            int ping        = g_pPlayerResource->GetPing(g_IEngine->GetLocalPlayer());
+            if (*force_ping <= ping)
+            {
+                cmdrate->m_fMaxVal = 999999999.9f;
+                cmdrate->m_fMinVal = -999999999.9f;
+                cmdrate->SetValue(-1);
+            }
+            else
+                cmdrate->SetValue(oldCmdRate);
```
TotallyNotElite, line 283:
```suggestion
            if (*force_ping <= ping && cmdrate->GetInt() != -1)
            {
                oldCmdRate = cmdrate->GetInt();
                cmdrate->m_fMaxVal = 999999999.9f;
                cmdrate->m_fMinVal = -999999999.9f;
                cmdrate->SetValue(-1);
            }
            else if (*force_ping > ping)
                cmdrate->SetValue(oldCmdRate);
```

BenCat07:
> @BenCat07 is the usage of cvars in the init routines without a nullptr check a potential concern?

Yeah bots with preload may just like crash

PR #1011, opened 2020-05-15
Fix and improve nightmode (see description)

BenCat07:
This commit first of all fixes nightmode, then second of all makes it more customizeable, and the third point is that it fixes antiafk

image
image
image

Note: There is currently a bug where it will spam your console if you change settings, this is avoidable through some game interfering methods, so i am waiting for feedback from @TotallyNotElite on this matter.

### DIFF
--- a/src/hooks/visual/FrameStageNotify.cpp
+++ b/src/hooks/visual/FrameStageNotify.cpp
@@ -17,11 +17,8 @@ static settings::Rgba nightmode_gui_color{ "visual.night-mode.gui-color", "00000
 static settings::Rgba nightmode_world_color{ "visual.night-mode.world-color", "000000FF" };
 static settings::Boolean no_shake{ "visual.no-shake", "true" };
 
-// Previous values
-static float old_nightmode_world{ 0.0f };
-static float old_nightmode_gui{ 0.0f };
-static rgba_t old_nightmode_world_color(0.0f, 0.0f, 0.0f, 1.0f);
-static rgba_t old_nightmode_gui_color(0.0f, 0.0f, 0.0f, 1.0f);
+// Should we update nightmode?
+static bool update_nightmode = false;
 
 // Which strings trigger this nightmode option
 std::vector<std::string> world_strings = { "SkyBox", "World" };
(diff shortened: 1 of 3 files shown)


src/hooks/visual/FrameStageNotify.cpp
```diff
@@ -26,7 +37,7 @@ DEFINE_HOOKED_METHOD(FrameStageNotify, void, void *this_, ClientFrameStage_t sta
 
     PROF_SECTION(FrameStageNotify_TOTAL);
 
-    if (old_nightmode != *nightmode)
+    if (old_nightmode_gui != *nightmode_gui || old_nightmode_world != *nightmode_world || old_nightmode_world_color != *nightmode_world_color || old_nightmode_gui_color != *nightmode_gui_color)
```
TotallyNotElite, line 40: Learn to use a~~n~~ ~~egg~~ bool. And callbacks.

src/hooks/visual/FrameStageNotify.cpp
```diff
@@ -42,20 +53,48 @@ DEFINE_HOOKED_METHOD(FrameStageNotify, void, void *this_, ClientFrameStage_t sta
...
+                {
                     pMaterial->ColorModulate(1.0f, 1.0f, 1.0f);
+                }
             }
         }
-        old_nightmode = *nightmode;
+        old_nightmode_gui         = *nightmode_gui;
+        old_nightmode_gui_color   = *nightmode_gui_color;
+        old_nightmode_world       = *nightmode_world;
+        old_nightmode_world_color = *nightmode_world_color;
```
TotallyNotElite, line 97: no.

src/hooks/visual/FrameStageNotify.cpp
BenCat07: Callbacks don't work in this case, it needs to be done in this exact loop, But yeah i'll make it look nicer

src/hooks/visual/FrameStageNotify.cpp
BenCat07: see above

PR #1019, opened 2020-05-20
Sentence Capitalization

LightPower392:
Capitalization changed to sentence case.


LightPower392:
Two additional notes for tooltips: tooltips set in the XML file cause the slider's value to not show up, and elements of a list can not show a tooltip.

BenCat07:
Yeah, both of these are a bug, so we shouls avoid sliders with tooltips or merge their tooltips.

TotallyNotElite:
Do you mind removing the merge commits? You can rebase instead.

PR #1020, opened 2020-05-21
Improve navigation

BenCat07:
Make navbot take more optimal routes and generally not run into walls/areas it can't go as often.

- Don't mark World as Potentially Passable
- Don't walk into enemy spawn
- Don't try to exit spawn as blue in setup time
- Don't run to other, blocked sections of the map

Also fix crouchjumping.


src/hack.cpp
```diff
@@ -336,6 +337,7 @@ free(logname);*/
     InitNetVars();
     g_pLocalPlayer    = new LocalPlayer();
     g_pPlayerResource = new TFPlayerResource();
+    g_pTeamRoundTimer = new CTeamRoundTimer();
```
TotallyNotElite, line 340: This causes a memory leak. Too bad!

src/hooks/CreateMove.cpp
```diff
@@ -242,6 +243,10 @@ DEFINE_HOOKED_METHOD(CreateMove, bool, void *this_, float input_sample_time, CUs
         PROF_SECTION(CM_PlayerResource);
         g_pPlayerResource->Update();
     }
+    {
+        PROF_SECTION(CM_TeamTimer);
+        g_pTeamRoundTimer->Update();
+    }
```
TotallyNotElite, line 249: Very early event callback :thinking: Any reason not to use one?

src/navparser.cpp
```diff
@@ -95,24 +97,31 @@ float getZBetweenAreas(CNavArea *start, CNavArea *end)
...
+    ignore_status status = vischeck_failed;
+    // Is world blocking it?
+    if (IsVectorVisibleNavigation(first, second, LOCAL_E, MASK_PLAYERSOLID))
+    {
+        status = vischeck_success;
+        // Is something else blocking it?
+        if (!IsVectorVisible(first, second, true, LOCAL_E, MASK_PLAYERSOLID))
+            status = vischeck_blockedentity;
+    }
+    return status;
```
TotallyNotElite, line 115: What even is this status nonsense?
```suggestion
    // Is world blocking it?
    if (IsVectorVisibleNavigation(first, second, LOCAL_E, MASK_PLAYERSOLID))
    {
        // Is something else blocking it?
        if (!IsVectorVisible(first, second, true, LOCAL_E, MASK_PLAYERSOLID))
            return vischeck_blockedentity;
        else
            return vischeck_success;
    }
    return vischeck_failed;
```

src/navparser.cpp
```diff
@@ -676,12 +696,21 @@ static void cm()
...
-            current_user_cmd->buttons |= (IN_JUMP | IN_DUCK);
+        if (!local || !(local->m_attributeFlags & (NAV_MESH_NO_JUMP | NAV_MESH_STAIRS)))
+        {
+            static bool flip_action = false;
+            // Make it crouch the second tick
+            current_user_cmd->buttons |= flip_action ? IN_DUCK : IN_JUMP;
+
+            // Update jump timer now
+            if (flip_action)
+                last_jump.test_and_set(200);
```
TotallyNotElite, line 711: Didn't you literally say "update" jump timer, yet you don't use .update xddd

src/trace.cpp
```diff
@@ -108,6 +108,60 @@ TraceType_t trace::FilterNoPlayer::GetTraceType() const
...
+
+void trace::FilterNavigation::SetSelf(IClientEntity *self)
+{
+    if (self == nullptr)
+    {
+        logging::Info("nullptr in FilterNavigation::SetSelf");
+        return;
+    }
+    m_pSelf = self;
+}
```
TotallyNotElite, line 129: m_pSelf is never used.

src/hooks/CreateMove.cpp
BenCat07: i Used it just like player resource

src/navparser.cpp
BenCat07: i think my brain shut down there

src/trace.cpp
BenCat07: Hm true, i can remove it in this case

src/navparser.cpp
BenCat07: Oops

BenCat07: Fixed.

src/trace.cpp
BenCat07: Fixed.

src/hack.cpp
BenCat07: Fixed. i also delete g_pLocalPlayer and g_pPlayerResource on shutdown now.

PR #1029, opened 2020-05-23
Allow up to 255 bots

ItsiAdam:


src/ipc.cpp
```diff
@@ -133,7 +133,7 @@ CatCommand debug_get_ingame_ipc("ipc_debug_dump_server", "Show other bots on ser
     int count        = 0;
     unsigned highest = 0;
     std::vector<unsigned> botlist{};
-    for (unsigned i = 1; i < cat_ipc::max_peers; i++)
+    for (unsigned i = 0; i < cat_ipc::max_peers; i++)
```
TotallyNotElite, line 136: Don't include the world entity wtf

ItsiAdam: idk if you're joking but this is looping over the ipc peers, not entities

ItsiAdam:
requires nullworks/simple-ipc#5  nullworks/cathook-ipc-web-panel#11 and  nullworks/cathook-ipc-server#9

BenCat07:
I've looked at this before but we would have to merge all of them at once and update simple-ipc submodule paths too

BenCat07:
Just updated the submodule, so i will reapprove for @TotallyNotElite's time's sake

PR #1033, opened 2020-05-24
Abandon logic improvements

ItsiAdam:
Don't connect if catbots already in the server (reads abandon-if-bots-ipc) and don't count self when checking count_ipc

### DIFF
--- a/src/hacks/CatBot.cpp
+++ b/src/hacks/CatBot.cpp
@@ -20,7 +20,7 @@ namespace hacks::shared::catbot
 {
 static settings::Boolean auto_disguise{ "misc.autodisguise", "true" };
 
-static settings::Int abandon_if_ipc_bots_gte{ "cat-bot.abandon-if.ipc-bots-gte", "0" };
+settings::Int abandon_if_ipc_bots_gte{ "cat-bot.abandon-if.ipc-bots-gte", "0" };
 static settings::Int abandon_if_humans_lte{ "cat-bot.abandon-if.humans-lte", "0" };
 static settings::Int abandon_if_players_lte{ "cat-bot.abandon-if.players-lte", "0" };
 
--- a/src/hacks/CatBot.cpp
+++ b/src/hacks/CatBot.cpp
@@ -788,8 +788,11 @@ void update()
         ipc_list.clear();
         int count_total = 0;
 
-        for (int i = 1; i <= g_IEngine->GetMaxClients(); ++i)
+        for (int i = 0; i <= g_IEngine->GetMaxClients(); ++i)
         {
+            if (g_IEngine->GetLocalPlayer() == i)
+                continue;
+
             if (g_IEntityList->GetClientEntity(i))
                 ++count_total;
             else
--- a/src/hooks/ClientCmd_Unrestricted.cpp
+++ b/src/hooks/ClientCmd_Unrestricted.cpp
@@ -8,10 +8,10 @@ namespace hooked_methods
 
 DEFINE_HOOKED_METHOD(ClientCmd_Unrestricted, void, IVEngineClient013* _this, const char* command)
 {
-    if (hacks::shared::catbot::abandon_if_ipc_bots_gte) {
+    if (ipc::peer && hacks::shared::catbot::abandon_if_ipc_bots_gte) {
         std::string command_str(command);
         
-        if (ipc::peer && command_str.length() >= 20 && command_str.substr(0, 7) == "connect") {
+        if (command_str.length() >= 20 && command_str.substr(0, 7) == "connect") {
             unsigned count_ipc = 0;
 
             if (command_str.substr(command_str.length() - 11, 11) == "matchmaking") {
--- a/src/hooks/ClientCmd_Unrestricted.cpp
+++ b/src/hooks/ClientCmd_Unrestricted.cpp
@@ -1,0 +1,44 @@
+#include "HookedMethods.hpp"
+#include "MiscTemporary.hpp"
+#include "CatBot.hpp"
+#include "interfaces.hpp"
+#include "ipc.hpp"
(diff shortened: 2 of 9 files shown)


src/hacks/CatBot.cpp
```diff
@@ -788,8 +788,11 @@ void update()
         ipc_list.clear();
         int count_total = 0;
 
-        for (int i = 1; i <= g_IEngine->GetMaxClients(); ++i)
+        for (int i = 0; i <= g_IEngine->GetMaxClients(); ++i)
         {
+            if (g_IEngine->GetLocalPlayer() == i)
+                continue;
```
TotallyNotElite, line 794: I don't think it's a good idea to change this behavior, since this will basically break existing configs.

src/hacks/CatBot.cpp
```diff
@@ -788,8 +788,11 @@ void update()
         ipc_list.clear();
         int count_total = 0;
 
-        for (int i = 1; i <= g_IEngine->GetMaxClients(); ++i)
+        for (int i = 0; i <= g_IEngine->GetMaxClients(); ++i)
```
TotallyNotElite, line 791: Why are you including the world entity?

src/hooks/ClientCmd_Unrestricted.cpp
```diff
@@ -0,0 +1,44 @@
...
+DEFINE_HOOKED_METHOD(ClientCmd_Unrestricted, void, IVEngineClient013* _this, const char* command)
+{
+    if (ipc::peer && hacks::shared::catbot::abandon_if_ipc_bots_gte) {
+        std::string command_str(command);
+        
+        if (command_str.length() >= 20 && command_str.substr(0, 7) == "connect") {
+            unsigned count_ipc = 0;
+
+            if (command_str.substr(command_str.length() - 11, 11) == "matchmaking") {
+                std::string server_ip = command_str.substr(8, command_str.length() - 20);
```
TotallyNotElite, line 18: You should probably just split the string at spaces instead of whatever you are doing here.

src/hooks/ClientCmd_Unrestricted.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+                auto &peer_mem = ipc::peer->memory;
+
+                for (unsigned i = 0; i < cat_ipc::max_peers; i++) {
+                    if (i != ipc::peer->client_id && !peer_mem->peer_data[i].free && peer_mem->peer_user_data[i].connected && peer_mem->peer_user_data[i].ingame.server) {
+
+                        std::string remote_server_ip(peer_mem->peer_user_data[i].ingame.server);
+                        if (remote_server_ip.compare(server_ip) == 0) {
```
TotallyNotElite, line 26: Use operator == instead of .compare.

src/hooks/ClientCmd_Unrestricted.cpp
```diff
@@ -0,0 +1,44 @@
...
+                            count_ipc++;
+                        }
+                    }
+                }
+            }
+
+            if (count_ipc > 0 && count_ipc >= (int)hacks::shared::catbot::abandon_if_ipc_bots_gte)
+            {
+                // if (hacks::shared::catbot::auto-requeue)
+                tfmm::startQueue();
```
TotallyNotElite, line 36: This check is unnecessary. The requeue process will be started automatically if it is enabled.

src/hacks/CatBot.cpp
ItsiAdam: Yeah but the current config just doesn't work, if you set it to 1 it'll just leave immediately.

src/hooks/ClientCmd_Unrestricted.cpp
ItsiAdam: potentially but this works fine

src/hacks/CatBot.cpp
TotallyNotElite: This is intended behavior. If you change this now, existing configs will be wrong.

Codefreak64:
Why is this still being refined, do you know how much pain us normal players have to go through, bot many bots we have to kick? and for what, to get a cheap laugh? if you want TF2 to DIE! then by all means please continue, but i'm not going to let a whole bunch of salty players try to KILL A GOOD GAME!!

ItsiAdam:
Mad? Kidiot? Kidiot mad?

ItsiAdam:
That is kid and idiot

src/hacks/CatBot.cpp
TotallyNotElite: I'll probably accept this anyway. It makes more sense this way.

src/hooks/ClientCmd_Unrestricted.cpp
BenCat07: Atleast put a comment to indicate how it does it

leKamikaze1:
> Why is this still being refined, do you know how much pain us normal players have to go through, bot many bots we have to kick? and for what, to get a cheap laugh? if you want TF2 to DIE! then by all means please continue, but i'm not going to let a whole bunch of salty players try to KILL A GOOD GAME!!

LMFAOOOOO

iloev:
> Why is this still being refined, do you know how much pain us normal players have to go through, bot many bots we have to kick? and for what, to get a cheap laugh? if you want TF2 to DIE! then by all means please continue, but i'm not going to let a whole bunch of salty players try to KILL A GOOD GAME!!

you do penny trading?
if yes then...
penny trading your mom
:trollface:

src/hooks/ClientCmd_Unrestricted.cpp
ItsiAdam: sugma balls

PR #1034, opened 2020-05-24
Look at healing target

ItsiAdam:


ItsiAdam:
Actually, this pr still needs improvements. If the intended target isn't the actual target, the bot should move around the actual target so we can aim at the intended target.

TotallyNotElite:
> the bot should move around the actual target

This is out of scope for autoheal.

PR #1043, opened 2020-05-27
Fix crashes after detaching with uncleared event callbacks

BenCat07:


BenCat07:
TL;DR:
haha g_pLocalPlayer (nullptr) go ->entity

PR #1046, opened 2020-05-29
Misc improvements

BenCat07:
This just has a few miscelanous improvements and fixes to existing things.

- Rework Slow aim (Previous one had major issues due to the way it worked
- Fix the Critmeter Text not being properly alligned
- Improve The chat censor filter
- Fix desyncs with crithack

### DIFF
--- a/src/hacks/AntiAntiAim.cpp
+++ b/src/hacks/AntiAntiAim.cpp
@@ -12,6 +12,7 @@ static settings::Boolean enable{ "anti-anti-aim.enable", "false" };
 static settings::Boolean debug{ "anti-anti-aim.debug.enable", "false" };
 
 std::unordered_map<unsigned, brutedata> resolver_map;
+std::array<CachedEntity *, 32> sniperdot_array;
 
 static inline void modifyAnlges()
 {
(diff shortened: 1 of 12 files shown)


src/hacks/AntiAntiAim.cpp
```diff
@@ -64,17 +64,64 @@ static float resolveAngleYaw(float angle, brutedata &brute)
...
+                CachedEntity *dot_ent = ENTITY(i);
+                // Not a sniper dot
+                if (CE_BAD(dot_ent) || dot_ent->m_iClassID() != CL_CLASS(CSniperDot))
+                    continue;
+                // Not the sniperdot of that entity
+                if (HandleToIDX(CE_INT(dot_ent, netvar.m_hOwnerEntity)) != ent->m_IDX)
+                    continue;
+                // Found sniper dot
+                sniper_dot = dot_ent;
+            }
```
TotallyNotElite, line 96: What about doing this once per CM and storing that in an array? This seems very inefficient to me.

src/hooks/DispatchUserMessage.cpp
```diff
@@ -215,7 +215,7 @@ DEFINE_HOOKED_METHOD(DispatchUserMessage, bool, void *this_, int type, bf_read &
                 break;
             }
 
-            std::vector<std::string> res = { "skid", "script", "cheat", "hak", "hac", "f1", "hax", "vac", "ban", "bot", "report", "kick" };
+            std::vector<std::string> res = { "skid", "script", "cheat", "hak", "hac", "f1", "hax", "vac", "ban", "bot", "report", "kick", "h@x", "hcak", "chaet", "one" };
```
TotallyNotElite, line 218: Why did you add h@x when the "@" is already replaced.

src/hacks/AntiAntiAim.cpp
BenCat07: Hm true i should store all sniper dots first, then use them

src/hooks/DispatchUserMessage.cpp
BenCat07: Will fix

PR #1048, opened 2020-05-30
Fix hitbox interpolation and fix punch angles in aiming functions

BenCat07:
Ayyyy


src/hacks/Misc.cpp
```diff
@@ -875,7 +875,7 @@ void Shutdown()
     ScoreboardColoring::patch_scoreboardcolor2->Shutdown();
 #endif
 }
-
+ProxyFnHook punchhook;
```
TotallyNotElite, line 878: I don't really think so

TotallyNotElite:
Where was this 3 years ago??

BenCat07:
> Where was this 3 years ago??

in development, we hope it was worth the wait.

PR #1064, opened 2020-06-12
Fix issues related to traces and Contracker

BenCat07:
The Contracker entirerly breaks animations, so i fixed that. Alongside that i also made melee traces use the proper mode and make snipers able to shoot through teammates.

### DIFF
--- a/src/hacks/Misc.cpp
+++ b/src/hacks/Misc.cpp
@@ -48,6 +48,12 @@ static settings::Boolean misc_drawhitboxes{ "misc.draw-hitboxes", "false" };
 static settings::Boolean misc_drawhitboxes_dead{ "misc.draw-hitboxes.dead-players", "false" };
 #endif
 
+#if ENABLE_TEXTMODE
+static settings::Boolean fix_cyoaanim{ "remove.contracker", "true" };
+#else
+static settings::Boolean fix_cyoaanim{ "remove.contracker", "false" };
+#endif
+
 #if !ENFORCE_STREAM_SAFETY && ENABLE_VISUALS
 static void tryPatchLocalPlayerShouldDraw(bool after)
 {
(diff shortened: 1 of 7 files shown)


src/trace.cpp
```diff
@@ -46,6 +46,34 @@ bool trace::FilterDefault::ShouldHitEntity(IHandleEntity *handle, int mask)
...
     case CL_CLASS(CFuncAreaPortalWindow):
         return false;
+    // Sniper rifle teammate check
+    case CL_CLASS(CTFPlayer):
+    {
+        // Check if self is set
+        if (m_pSelf)
+        {
+            // If what we hit is an enemy it does not matter
+            if (entity && CE_VALID(ENTITY(entity->entindex())) && !ENTITY(entity->entindex())->m_bEnemy())
```
TotallyNotElite, line 56: m_bEnemy does not take the team m_pSelf is in into consideration. This code expects m_pSelf to always be the local player, which is incorrect.

src/hacks/Misc.cpp
```diff
@@ -876,7 +882,24 @@ void Shutdown()
...
+    int *value_out = (int *) out;
+    if (!fix_cyoaanim)
+    {
+        *value_out = value;
+        return;
+    }
+    // Mark as not doing cyoa thing
+    if (CE_BAD(LOCAL_E) || pPlayer != RAW_ENT(LOCAL_E))
+        *value_out = false;
+}
```
TotallyNotElite, line 899: I mean I guess this works, but the best solution would be to just fix it on the cheat hitbox level instead of disabling the animation entirely.

src/trace.cpp
BenCat07: Oh, you are right

src/hacks/Misc.cpp
BenCat07: The issue is doing just that, because you would have to apply the hitboxes given at a specific animation, i don't see a real way to make that a thing as of right now

PR #1067, opened 2020-06-22
Fix the ping reducer changing clientside cmdrate

BenCat07:
This just sends the server your cmdrate instead of setting it clientside


BenCat07:
Done

PR #1078, opened 2020-07-30
Use a better and faster method of authentication

BenCat07:
This method can be spammed as much as we want which means we can send it like every 30 seconds instead of having to wait, also means bots can instantly reply to any auth

### DIFF
--- a/src/hooks/SendNetMsg.cpp
+++ b/src/hooks/SendNetMsg.cpp
@@ -16,8 +16,9 @@ static settings::Int newlines_msg{ "chat.prefix-newlines", "0" };
 static settings::Boolean log_sent{ "debug.log-sent-chat", "false" };
 static settings::Boolean answerIdentify{ "chat.identify.answer", "true" };
 static Timer identify_timer{};
-constexpr int CAT_IDENTIFY = 0xCA7;
-constexpr int CAT_REPLY    = 0xCA8;
+constexpr int CAT_IDENTIFY   = 0xCA7;
+constexpr int CAT_REPLY      = 0xCA8;
+constexpr float AUTH_MESSAGE = 1234567.0f;
 
 namespace hacks::shared::catbot
 {
@@ -26,8 +27,9 @@ void SendNetMsg(INetMessage &msg);
 namespace hooked_methods
 {
 
-static bool queue_ca8{};
-static Timer queue_ca8_timer{};
+static bool send_achievement_reply{};
+static bool send_drawline_reply{};
+static Timer send_achievement_reply_timer{};
 static Timer reply_timer{};
 
 void sendAchievementKv(int value)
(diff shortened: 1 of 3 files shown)


src/hooks/SendNetMsg.cpp
```diff
@@ -155,38 +198,46 @@ class AchievementListener : public IGameEventListener2
...
-    EC::Register(
-        EC::Shutdown, []() { g_IEventManager2->RemoveListener(&event_listener); }, "shutdown_event");
+    g_IEventManager2->AddListener(&event_listener, "cl_drawline", false);
     EC::Register(
         EC::LevelInit,
-        []() { // This will make the timer get triggered next test_and_set call
-            identify_timer.last -= std::chrono::minutes(5);
+        []() {
+            // Remove 1 minute from timer so it triggers on connect
+            identify_timer.last -= std::chrono::minutes(1);
```
TotallyNotElite, line 236: This is a terrible hack. Can you don't?

src/hooks/SendNetMsg.cpp
```diff
@@ -37,9 +37,52 @@ void sendAchievementKv(int value)
...
+    int id       = reinterpret_cast<int &>(x_value);
+
+    if (player_idx != 0xDEAD && (id == CAT_IDENTIFY || id == CAT_REPLY))
+    {
+        // Always reply and set on CA7 and only set on CA8
+        bool reply = id == CAT_IDENTIFY;
+        player_info_s info;
+        if (!g_IEngine->GetPlayerInfo(player_idx, &info))
+            return;
+        if (reply && *answerIdentify && player_idx != g_pLocalPlayer->entity_idx && reply_timer.test_and_set(2000))
```
TotallyNotElite, line 81: Edge case: What if an event is sent while this timer has not passed 2 seconds?

BenCat07: Hm true but we also want to avoid spam

src/hooks/SendNetMsg.cpp
BenCat07: ok

src/hooks/SendNetMsg.cpp
```diff
@@ -155,38 +208,44 @@ class AchievementListener : public IGameEventListener2
...
+            if (!*identify || CE_BAD(LOCAL_E) || !identify_timer.test_and_set(1000 * 30))
                 return;
-            sendIdentifyMessage(false);
+            sendIdentifyMessage(false, false);
         },
         "sendnetmsg_createmove");
     g_IEventManager2->AddListener(&event_listener, "achievement_earned", false);
+    g_IEventManager2->AddListener(&event_listener, "cl_drawline", false);
     EC::Register(
         EC::Shutdown, []() { g_IEventManager2->RemoveListener(&event_listener); }, "shutdown_event");
```
TotallyNotElite, line 248: Does this correctly remove both listeners?

src/hooks/SendNetMsg.cpp
```diff
@@ -155,38 +208,44 @@ class AchievementListener : public IGameEventListener2
...
+                sendIdentifyMessage(true, false);
                 queue_ca8 = false;
             }
-            // 2 minutes between each identify seems ok?
-            if (!*identify || CE_BAD(LOCAL_E) || !identify_timer.test_and_set(1000 * 60 * 2))
+            // Wait 30 seconds between identifies
+            if (!*identify || CE_BAD(LOCAL_E) || !identify_timer.test_and_set(1000 * 30))
                 return;
-            sendIdentifyMessage(false);
+            sendIdentifyMessage(false, false);
```
TotallyNotElite, line 242: So legacy identifies are no longer being sent?

src/hooks/SendNetMsg.cpp
BenCat07: Yes

src/hooks/SendNetMsg.cpp
BenCat07: Yep, only replies

BenCat07: It will now send one legacy identify on join

Codefreak64:
Why is this still being refined, do you know how much pain us normal players have to go through, bot many bots we have to kick? and for what, to get a cheap laugh? if you want TF2 to DIE! then by all means please continue, but i'm not going to let a whole bunch of salty players try to KILL A GOOD GAME!!

TotallyNotElite:
> if you want TF2 to DIE! then by all means please continue, but i'm not going to let a whole bunch of salty players try to KILL A GOOD GAME!!

If valve actually cared, maybe it wouldn't be dying. Many players gave up on the game a long time ago.

> Why is this still being refined

Cause it's a hobby.

> do you know how much pain us normal players have to go through, bot many bots we have to kick?

Wow, I really don't care. Complain to bot hosters, not devs.

PR #1080, opened 2020-08-05
Fix a crash caused by warp on hit

BenCat07:
Sometimes you can get hit by someone who is dormant/invalid or the event is so delayed that it arrives once you/your target are dead, this aims to fix that crash (obversed on the bots of a hoster)

### DIFF
--- a/src/hacks/Warp.cpp
+++ b/src/hacks/Warp.cpp
@@ -511,8 +511,9 @@ class WarpHurtListener : public IGameEventListener2
 {
 public:
     virtual void FireGameEvent(IGameEvent *event)
-    { // Not enabled
-        if (!enabled || !warp_on_damage)
+    {
+        // Not enabled
+        if (!isHackActive() || !enabled || !warp_on_damage)
             return;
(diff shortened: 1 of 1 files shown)


src/hacks/Warp.cpp
```diff
@@ -529,9 +530,20 @@ class WarpHurtListener : public IGameEventListener2
         // Check if victim is local player
         if (g_IEngine->GetPlayerForUserID(victim) != g_pLocalPlayer->entity_idx)
             return;
-        // Ignore projectiles for now
-        if (CE_VALID(ENTITY(attacker)) && GetWeaponMode(ENTITY(attacker)) == weapon_projectile)
+
+        // Check if the entities are alive and valid
+        CachedEntity *att = ENTITY(attacker);
```
TotallyNotElite, line 535: Why can you not use LOCAL_E here? This variable wouldn't need to exist at all, really

BenCat07: Victim is us, not attacker, but yeah fair point

PR #1090, opened 2020-08-17
Fix inconsistency and speed up identify

aUniqueUser:


BenCat07:
Panel being 0 is a mistake

BenCat07 (approved):
Fixed the delay being a bit too short

PR #1109, opened 2020-09-16
Improve textmode performance

BenCat07:
- Fully null Material Swap buffers
- Make map loading times shorter and take less cpu time by skipping ressource checks
- Remove CViewRender::Render calls from game
- Remove a sleep(15000) that gets used on instances marked as "textmode"


TotallyNotElite:
> sleep(15000)
sleep 250 hours :open_mouth:

BenCat07:
usleep(15000), my bad

TotallyNotElite (approved):

BenCat07:
Let me also ammend a comment real quick i missed on one bytepatch

TotallyNotElite (approved):

PR #1114, opened 2020-10-02
Fix issues caused by tf2 update

BenCat07:


include/helpers.hpp
```diff
@@ -226,3 +226,84 @@ template <typename... Args> std::string format(const Args &... args)
...
+    }
+};
+
+inline WeaponData_t *GetWeaponData(IClientEntity *weapon)
+{
+    weapon_info info(weapon);
+    int WeaponData = info.weapon_data;
+    int WeaponMode = info.weapon_mode;
+    return (WeaponData_t *) (WeaponData + sizeof(WeaponData_t) * WeaponMode + 1784);
+}
```
TotallyNotElite, line 309: I don't think this belong here. Why don't you make a new header for this?

TotallyNotElite (changes requested):

include/helpers.hpp
BenCat07: Sure

include/common.hpp
```diff
@@ -76,6 +76,7 @@
 #include <CNavFile.h>
 #include "HookTools.hpp"
 #include "bytepatch.hpp"
+#include "WeaponData.hpp"
```
TotallyNotElite, line 79: I REALLY don't think its necessary for every translation unit to have this

TotallyNotElite (changes requested):

include/common.hpp
BenCat07: Resolved

TotallyNotElite (approved):
Assuming it's tested

BenCat07:
> Assuming it's tested

indeed

PR #1130, opened 2020-10-17
Add sapper as pda

BenCat07:
Why is the sapper not using the sapper weapon class, this makes no sense.


TotallyNotElite (approved):
g_pLocalPlayer->holding_sapper exists too

BenCat07:
Yeah this is still dumb tho

PR #1134, opened 2020-10-19
Fix: Autoitem: Use maps instead of looping thru achivement items

TotallyNotElite:
Money worm

Money worm

This is not tested, if you accept this and it doesn't work it's your fault.

Money worm

@aUniqueUser is the MVP


aUniqueUser:
worm rich

BenCat07 (approved):
confirmed working

PR #1135, opened 2020-10-19
Make randoms not depend on rand()

aUniqueUser:
Changed UniformRandomInt and RandFloatRange to not depend on rand(), because this can cause little to no randomness depending on system.

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -962,18 +962,17 @@ bool AmbassadorCanHeadshot()
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


BenCat07 (approved):
confirmed working

TotallyNotElite (changes requested):
1. why have 2 random devices
2. why have two mt19937's
3. thread safety out the window?

TotallyNotElite (approved):

PR #1137, opened 2020-10-20
Fix truce and enable aimbot to aim at npcs

BenCat07:
This simply makes aimbot able to aim at npcs and the cheat will ignore players in truce.

- [ ] Request to update stable aswell


src/PlayerTools.cpp
```diff
@@ -27,6 +27,9 @@ static CatCommand forgive_all("pt_forgive_all", "Clear betrayal list", []() { be
 
 bool shouldTargetSteamId(unsigned id)
 {
+    // Don't shoot players in truce
+    if (isTruce())
+        return false;
```
TotallyNotElite, line 32: I think this should be in shouldTarget instead of shouldTargetSteamId

src/hacks/Aimbot.cpp
```diff
@@ -803,9 +803,39 @@ bool IsTargetStateGood(CachedEntity *entity)
...
+        // Cannot shoot this
+        if (entity->m_iTeam() == LOCAL_E->m_iTeam())
+            return false;
+
+        // Distance
+        if (EffectiveTargetingRange())
+        {
+            if (entity->m_flDistance() > (int) EffectiveTargetingRange())
+                return false;
+        }
```
TotallyNotElite, line 827: Does this code really need to be duplicated?

TotallyNotElite (changes requested):
Is there really no better way to determine truce?

src/hacks/Aimbot.cpp
BenCat07: talked out in dms, bt specific code exists

BenCat07:
We have no other way to determine a truce.

BenCat07:
the thing meant to tell us if there is a truce is broken and does not get networked

src/hacks/ESP.cpp
```diff
@@ -1106,16 +1117,34 @@ void _FASTCALL ProcessEntity(CachedEntity *ent)
     }
 
     int itemtype = ent->m_ItemType();
-    // Tank esp
-    if (classid == CL_CLASS(CTFTankBoss) && tank)
+    // NPC esp
+    if (classid == CL_CLASS(CTFTankBoss) || classid == CL_CLASS(CMerasmus) || classid == CL_CLASS(CMerasmusDancer) || classid == CL_CLASS(CZombie) || classid == CL_CLASS(CEyeballBoss) || classid == CL_CLASS(CHeadlessHatman) && npc)
```
TotallyNotElite, line 1121: This looks like duplicated code with Aimbot? Npc bool should also be checked first.

src/hacks/ESP.cpp
```diff
@@ -1106,16 +1117,34 @@ void _FASTCALL ProcessEntity(CachedEntity *ent)
...
+        if (classid == CL_CLASS(CTFTankBoss))
+            AddEntityString(ent, tank_str, colors::FromRGBA8(0, 128, 0, 255));
+        if (classid == CL_CLASS(CMerasmus) || classid == CL_CLASS(CMerasmusDancer))
+            AddEntityString(ent, merasmus_str, colors::FromRGBA8(0, 128, 0, 255));
+        if (classid == CL_CLASS(CZombie))
+            AddEntityString(ent, skeleton_str, colors::FromRGBA8(0, 128, 0, 255));
+        if (classid == CL_CLASS(CEyeballBoss))
+            AddEntityString(ent, monoculus_str, colors::FromRGBA8(0, 128, 0, 255));
+        if (classid == CL_CLASS(CHeadlessHatman))
+            AddEntityString(ent, horsemann_str, colors::FromRGBA8(0, 128, 0, 255));
```
TotallyNotElite, line 1132: Don't pull a Yanderedev. Switch case please!

src/hacks/ESP.cpp
```diff
@@ -1106,16 +1117,34 @@ void _FASTCALL ProcessEntity(CachedEntity *ent)
...
         // Dropped weapon esp
     }
     else if (classid == CL_CLASS(CTFDroppedWeapon) && item_esp && item_dropped_weapons)
     {
-        AddEntityString(ent, std::string("Dropped Weapon")); //I've never seen this not being "WEAPON CTFDroppedWeapon" so why not make it prettier?
+        AddEntityString(ent, std::string("Dropped Weapon"));
+
+        // Gargoyle esp
+    }
+    else if (classid == CL_CLASS(CHalloweenGiftPickup) && item_esp && item_gargoyle)
```
TotallyNotElite, line 1142: The item_esp bool is checked multiple times. Try to avoid this. Bool should also always be checked first, not the condition.

src/hacks/ESP.cpp
```diff
@@ -1133,6 +1162,12 @@ void _FASTCALL ProcessEntity(CachedEntity *ent)
             AddEntityString(ent, mvm_money_str);
         }
 
+        // Pumpkin bomb esp
+    }
+    else if (classid == CL_CLASS(CTFPumpkinBomb) && item_esp && item_pumpkinbomb)
```
TotallyNotElite, line 1167: The item_esp bool is checked multiple times. Try to avoid this. Bool should also always be checked first, not the condition.

src/itemtypes.cpp
```diff
@@ -71,7 +73,12 @@ ItemManager::ItemManager() : mapper()
     RegisterModelMapping("models/props_halloween/hwn_spellbook_upright_major.mdl", ITEM_SPELL_RARE);
     RegisterModelMapping("models/items/crystal_ball_pickup_major.mdl", ITEM_SPELL_RARE);
 
-    RegisterSpecialMapping([](CachedEntity *ent) -> bool { return ent->m_iClassID() == CL_CLASS(CTFAmmoPack); }, ITEM_AMMO_MEDIUM);
+    // == GHOSTS
+    RegisterModelMapping("models/props_halloween/ghost.mdl", HALLOWEEN_GHOST);
+    RegisterModelMapping("models/props_halloween/ghost_no_hat_red.mdl", HALLOWEEN_GHOST_NOHAT_RED);
+    RegisterModelMapping("models/props_halloween/ghost_no_hat.mdl", HALLOWEEN_GHOST_NOHAT);
+
+    RegisterSpecialMapping([](CachedEntity *ent) -> bool { return ent->m_iClassID() == CL_CLASS(CTFAmmoPack) && !ITEM_CRUMPKIN; }, ITEM_AMMO_MEDIUM);
```
TotallyNotElite, line 81: "!ITEM_CRUMPKIN" always resolves to false, therefore this lambda will always return false.

TotallyNotElite (changes requested):
@BenCat07 did you even look at this code

BenCat07:
instant regret not looking at the code as it was said to be working

src/hacks/ESP.cpp
BenCat07: Will fix

TotallyNotElite (approved):

PR #1144, opened 2020-10-27
Revving and taunting condition esp

NotAnUser1:
Taunting is for people like me who has the tauntfix (https://t.me/tauntfix_reborn) mod installed, what that mod does in simple words are removing taunting animations at an attempt of making hitboxes match with server

### DIFF
--- a/src/hacks/ESP.cpp
+++ b/src/hacks/ESP.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * HEsp.cpp
  *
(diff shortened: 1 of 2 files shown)


TotallyNotElite:
Not sure why you'd need the mod, especially because cathook already has a remove taunts feature

TotallyNotElite (approved):

BenCat07 (approved):
lhtm

PR #1150, opened 2020-10-31
Fix aiming at buildings

BenCat07:
Render bounds stretch under the floor for some reason, and apparently those npcs do have proper hitboxes we can aim at, so do it
image

^this is what it tried to aim at before, which occasionally (especially on teleporters) lead to alot of issues. It now aims at the middle most hitbox, which also makes aiming at sentries more reliable.

### DIFF
--- a/src/hacks/Aimbot.cpp
+++ b/src/hacks/Aimbot.cpp
@@ -1142,9 +1142,7 @@ const Vector &PredictEntity(CachedEntity *entity)
         // NPCs (Skeletons, merasmus, etc)
         else if (entity->m_Type() == ENTITY_NPC)
         {
-            auto mins = RAW_ENT(entity)->GetCollideable()->OBBMins();
-            auto maxs = RAW_ENT(entity)->GetCollideable()->OBBMaxs();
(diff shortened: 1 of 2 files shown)


BenCat07:
Also aims at hitboxes on npcs instead of their center

EngineerGaming69:
Amazing

EngineerGaming69 (approved):
Thx for the fix

TotallyNotElite (changes requested):
Consider this:
Hitbox count: 1
1 / 2 = 0.5 -> integer division in this implementation leads to floor -> 0
0 - 1 = -1 -> Out of bounds

EngineerGaming69 (dismissed):
Thx for the fix

BenCat07:
> Consider this:
> Hitbox count: 1
> 1 / 2 = 0.5 -> integer division in this implementation leads to floor -> 0
> 0 - 1 = -1 -> Out of bounds

i forgot it floors. my bad

EngineerGaming69:
will this only fix guns or will it fix melee too?

BenCat07:
It will fix both guns and melee

TotallyNotElite (approved):
LGTM

PR #1151, opened 2020-10-31
Fix building aim

BenCat07:
Soooo apparently the hitboxes are invalid on dormant entities... oops. (No changes in aimbot needed, as aimbot never aims at dormant ents)


TotallyNotElite (changes requested):
This executes GetHitboxes twice. Use a variable.

TotallyNotElite (approved):

PR #1163, opened 2020-11-13
new default dominate says

meowinghuman:
Saw a comment asking for new dominate says, well here ya go.

### DIFF
--- a/src/hacks/DominateSay.cpp
+++ b/src/hacks/DominateSay.cpp
@@ -18,10 +18,9 @@ const std::vector<std::string> builtin_default = {
     "%name%, nice skill, you sell?",
     "%name%, umad?",
     "smh i have %dominum% dominations",
-    "R.I.P %name%, they were fat, and died a virgin!",
     "Well, off to vist %name%'s mother!",
     "Cleanup on asile 3, %name% left their skill!",
-    "Congrats %name%! You just won 50 precent off a free asswhoopin! Call now and get another one absolutely free!",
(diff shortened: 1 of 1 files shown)


src/hacks/DominateSay.cpp
```diff
@@ -18,6 +18,10 @@ const std::vector<std::string> builtin_default = {
     "%name%, nice skill, you sell?",
     "%name%, umad?",
     "smh i have %dominum% dominations",
+    "R.I.P %name%, they were fat, and died a virgin!",
```
TotallyNotElite, line 21: I don't think this is github TOS compliant, we've been fck'd for killsays before:

image

src/hacks/DominateSay.cpp
```diff
@@ -18,6 +18,10 @@ const std::vector<std::string> builtin_default = {
...
+    "Well, off to vist %name%'s mother!",
+    "Cleanup on asile 3, %name% left their skill!",
+    "Congrats %name%! You just won 50 precent off a free asswhoopin! Call now and get another one absolutely free!",
```
TotallyNotElite, line 24: > You just won 50 precent off a **free**
You can't get 50 percent off something that is free

TotallyNotElite (changes requested):

src/hacks/DominateSay.cpp
G-4-M-3-R: > You can't get 50 percent off something that is free

That just makes it better

TotallyNotElite (approved):

PR #1164, opened 2020-11-13
Added spies next to spy count

meowinghuman:
Not really needed but I thought was somewhat cool.

### DIFF
--- a/src/hacks/SpyAlert.cpp
+++ b/src/hacks/SpyAlert.cpp
@@ -80,7 +80,7 @@ void Draw()
             }
             if (voicemenu && lastVoicemenu.test_and_set(5000))
                 g_IEngine->ClientCmd_Unrestricted("voicemenu 1 1");
-            AddCenterString(format("BACKSTAB WARNING! ", (int) (closest_spy_distance / 64 * 1.22f), "m (", spy_count, ")"), colors::red);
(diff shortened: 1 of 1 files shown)


TotallyNotElite:
@BenCat07 thoughts? I think this may be counterproductive because it may block up the center of the screen more.

BenCat07:
Yeah may be, but i've not tested it yet

src/hacks/SpyAlert.cpp
```diff
@@ -80,7 +80,7 @@ void Draw()
             }
             if (voicemenu && lastVoicemenu.test_and_set(5000))
                 g_IEngine->ClientCmd_Unrestricted("voicemenu 1 1");
-            AddCenterString(format("BACKSTAB WARNING! ", (int) (closest_spy_distance / 64 * 1.22f), "m (", spy_count, ")"), colors::red);
+            AddCenterString(format("BACKSTAB WARNING! ", (int) (closest_spy_distance / 64 * 1.22f), "m (", spy_count, " SPIES!", ")"), colors::red);
```
BenCat07, line 83: If you are going to add this, make it show SPY if there is a single one, and SPIES if there are multiple :)

BenCat07 (changes requested):

PR #1168, opened 2020-11-18
Add strafeprediction to cathook

TotallyNotElite:
It is time.

Best case scenario: https://t.me/nullworks/501090
Some gameplay: https://www.youtube.com/watch?v=iNI0qfqjPjg

As far as I'm aware, we are the only open source cheat with strafe prediction now.

You can mess around with debug.strafepred.samplesize and see what you like best, but 10 should be a reasonable default (movement in the last 10/66 of a second is used to calculate the next position).

Just FYI for airshots: Use static hitbox mode and choose something like spine 0. Gives the best results in my experience.


BenCat07 (approved):
LGTM

PR #1169, opened 2020-11-19
Make cathook identify-neutral by default

TotallyNotElite:
This will make cathook only send auth packets when
1. Someone else sent one first
2. We are in textmode (we are a bot)

This was added to prevent external cheats from taking advantage of cathook's auth features without returning the favour and to prevent cathook users from landing on bot blacklists.


TotallyNotElite:
Tested, works as advertised

BenCat07 (approved):
LGTM

PR #1176, opened 2020-11-24
Fix disguised spies being ignored, fix auth to rage users

TotallyNotElite:


src/hooks/SendNetMsg.cpp
```diff
@@ -80,7 +80,7 @@ void ProcessSendline(IGameEvent *kv)
             return;
         // CA7 = Reply and change state
         // CA8 = Change state
-        if (id == CAT_IDENTIFY && *answerIdentify && player_idx != g_pLocalPlayer->entity_idx)
```
BenCat07, line 83: Maybe we should generally not auth when a rage user is on the server, but not sure

BenCat07 (approved):
LGTM

PR #1183, opened 2020-11-28
Add autoparty

delimeats-ch:
cheemsburger gaming


delimeats-ch:
Untested, needs menu stuff still

src/hacks/AutoParty.cpp
```diff
@@ -0,0 +1,286 @@
...
+                    leave_party(client, false);
+                }
+            }
+        }
+    }
+}
+
+static InitRoutine init([]() {
+    host_list.installChangeCallback([](settings::VariableBase<std::string> &var, std::string after) { repopulate(); });
+    EC::Register(EC::CreateMove, party_routine, "cm_autoparty", EC::average);
```
TotallyNotElite, line 284: Createmove is not exactly ideal for this, because createmove **only works if you are in a game**. You should use EC::Paint instead (not to be confused with EC::Draw).

TotallyNotElite (changes requested):
Ok so a few more things:

- Instead of calling the engine for sending commands directly, you should use https://github.com/nullworks/cathook/blob/9c1b236a24db3a7324f5f4cbd374188efabd6e2a/src/hack.cpp#L105 instead.
- Instead of this "manual" configuration as provided by this module, what would you say about using the IPC for coordinating automatic leader selection, etc? Is that something you'd be interested in doing?

delimeats-ch:
>     * Instead of this "manual" configuration as provided by this module, what would you say about using the IPC for coordinating automatic leader selection, etc? Is that something you'd be interested in doing?

Not really, manual selection is a lot more versatile and not too hard to work with imo. Easy to make it work with premium bots or if you want to drag some bots around with you. I'll probably fix everything else.

TotallyNotElite:
Looks good :+1:! I'll accept and merge once it's tested and menu entries are added.

delimeats-ch:
Should be good to merge now

delimeats-ch:
One more thing to fix actually, setting party-hosts via `cat set` doesn't trigger the callback. Not sure how to fix that other than just loading it every time the routine runs, which feels kind of dumb.

BenCat07:
well setting it via cat set should just trigger the callback

delimeats-ch:
Mind testing? Doesn't work for me, but editing it from the menu works fine

BenCat07:
well if it didn't work then i'd be unable to enable/disable party bypas for example, but it works just fine soo

delimeats-ch:
Is the new value of a changed setting already reflected while the callback runs, or is the new value provided through the callback, and that needs to get passed in?

delimeats-ch:
e.g. `host_list.installChangeCallback([](settings::VariableBase<std::string> &var, std::string after) { repopulate(); });`
repopulate() assumes the new value is already in the setting

delimeats-ch:
I'll look tomorrow

BenCat07:
> Is the new value of a changed setting already reflected while the callback runs, or is the new value provided through the callback, and that needs to get passed in?

undefined behaviour, use the `after` variable on callback

TotallyNotElite:
Actually, the behaviour is defined. The variable will always contain the OLD value until all other callbacks have been called.

TotallyNotElite:
This PR has kind of gone out of scope and includes changes that should probably be a separate PR. Also, you need to use the "after" variable like @BenCat07 said.

delimeats-ch:
> This PR has kind of gone out of scope and includes changes that should probably be a separate PR. Also, you need to use the "after" variable like @BenCat07 said.

ha ha development branch go brr
Fixed, ready for review.

TotallyNotElite:
Those are a lot of XML changes. We should try to see what formatting for spaces does instead.

delimeats-ch:
> Those are a lot of XML changes. We should try to see what formatting for spaces does instead.

Converting to spaces is a much smaller commit, want me to do that instead?

TotallyNotElite:
Sounds good, thanks

delimeats-ch:
> Sounds good, thanks

Done

TotallyNotElite (approved):
@BenCat07 review and merge

TotallyNotElite:
Pinging @BenCat07 again

BenCat07 (approved):

TotallyNotElite:
:tada:

PR #1191, opened 2020-12-08
Rapidfire improvements

BenCat07:
- Add Instant zoom (and Heatmaker support)
- Fix misses while moving and using Rapidfire
- Improve Performance while using Rapidfire, alot

might want to squish these commits


src/core/interfaces.cpp
```diff
@@ -169,6 +170,7 @@ void CreateInterfaces()
     g_IStudioRender     = BruteforceInterface<IStudioRender>("VStudioRender", sharedobj::studiorender());
     g_IVRenderView      = BruteforceInterface<IVRenderView>("VEngineRenderView", sharedobj::engine());
     g_IMaterialSystemHL = (IMaterialSystem *) g_IMaterialSystem;
+    g_IToolFramework    = BruteforceInterface<IToolFrameworkInternal>("VTOOLFRAMEWORKVERSION", sharedobj::engine());
```
BenCat07, line 173: hey idiot, this is not visual only

delimeats-ch:
Very nice.

src/core/interfaces.cpp
TotallyNotElite: Uh

TotallyNotElite (dismissed):
IMO this entire approach is gay

src/core/interfaces.cpp
BenCat07: i need to fix this for textmode

PR #1192, opened 2020-12-09
Install data to "DataPath" as intended

notroboot:
Fixes a bug where `make install` would install to `/opt/cathook` regardless of what the the CMake var `DataPath` was set to.

### DIFF
--- a/install-data
+++ b/install-data
@@ -1,15 +1,18 @@
 #!/usr/bin/env bash
 
 [[ ! -z "$SUDO_USER" ]] && user=$SUDO_USER || user=$LOGNAME
-if ! [ -d "/opt/cathook/data" ]; then
-    echo "Creating cathook data directory at /opt/cathook/data"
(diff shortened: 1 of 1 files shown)


notroboot:
A little more explanation since the commit doesn't really explain itself:
https://github.com/nullworks/cathook/blob/13a0c43b4e5fc45328d045e5eff7a1fc38e35cb7/CMakeLists.txt#L222-L226
`install-data.sh` was supposed to accept an optional (?) command-line argument specifying the install destination for Cathook's data, but the script never implemented it and would always install to `/opt/cathook`.

https://github.com/nullworks/cathook/blob/13a0c43b4e5fc45328d045e5eff7a1fc38e35cb7/install-data#L3-L15

All this patch does is add that missing command-line argument, as originally intended.

TotallyNotElite (approved):
Cool thanks

PR #1299, opened 2020-12-31
Fix a autoitem and a chams bug

aUniqueUser:
Autoitem would not work with '-1' because it would try to get the item.
Chams would still apply local player cham with teammates on and local player off.


TotallyNotElite (approved):

TotallyNotElite:
LGTM, Rebase and Merge button is all yours!

PR #1304, opened 2021-01-01
Remove cotire & projectile silent

TotallyNotElite:
1. Ubuntu 18.04 users are rarer than ever before. Not worth it to keep deprecated software around. Cathook will still work for them, it will just compile slower.
2. Projectile silent was removed at the request of @BenCat07


BenCat07 (approved):
LGTM

PR #1313, opened 2021-01-06
get memory map without finding our pid first

Jan200101:
replaced the formatted string with a string literal for "/proc/self/maps" since it maps to the same thing


TotallyNotElite (approved):
Lgtm, didn't even know we did this

PR #1328, opened 2021-01-15
Add support for void linux

Roguezilla:


aUniqueUser (approved):
Works on my void machine

PR #1348, opened 2021-02-11
Revert "Remove stealth kill patch that causes crashes for some users"

TotallyNotElite:
Reverts nullworks/cathook#1340


BenCat07 (approved):
LGTM

PR #1342, opened 2021-02-09
Check if attach is ran as root + Add Debian support to dependencycheck script

Trapacid:
Adds Debian support for the dependency check script which was giving me issues using Debian.

PR #1349, opened 2021-02-11
Check if attach is ran as root + Add Debian support to dependencycheck script 

Trapacid:
[Old PR](https://github.com/nullworks/cathook/pull/1342)

(I don't use git that much :P)

### DIFF
--- a/attach
+++ b/attach
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+# Verify that we are root
+if [ ! $(id -u) -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


TotallyNotElite (approved):
LGTM

TotallyNotElite:
Mmmmm, CircleCI really does not like you.

TotallyNotElite:
Cathook-GUI will still consider debian as unsupported and will launch in binary mode instead of source mode.

This change will take effect once stable is updated.

PR #1351, opened 2021-02-13
Fix Absurdely janky projectile problems with initial velocity

BenCat07:


TotallyNotElite (dismissed):
trust.avi

aUniqueUser (approved):
lgtm

acidtripparty:
Why do we need separate calculation for initial velocity?

BenCat07:
because we need to compensate for the fact that some projectiles start with an upwards velocity already applied, if we do not compensate for that we miss alot of shots

acidtripparty:
Thanks for the explanation!

PR #1354, opened 2021-02-15
Add zoom key

BenCat07:
Just add a way to zoom in without a sniper scope, useful if you want to shoot someone from far away with a projectile for example, or just to check out what's going on in that direction.

### DIFF
--- a/src/hooks/visual/OverrideView.cpp
+++ b/src/hooks/visual/OverrideView.cpp
@@ -10,7 +10,10 @@
 static settings::Float override_fov{ "visual.fov", "0" };
 static settings::Float freecam_speed{ "visual.freecam-speed", "800.0f" };
 static settings::Button freecam{ "visual.freecam-button", "<none>" };
+static settings::Button zoom_key{ "visual.zoom-key", "<null>" };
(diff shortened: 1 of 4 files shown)


src/hooks/visual/OverrideView.cpp
```diff
@@ -32,6 +34,15 @@ DEFINE_HOOKED_METHOD(OverrideView, void, void *this_, CViewSetup *setup)
...
     }
+    if (zoom_key && zoom_key.isKeyDown())
+    {
+        auto default_fov               = g_ICvar->FindVar("default_fov");
+        float sens_factor              = (*zoom_fov / setup->fov) * (setup->fov / default_fov->GetFloat());
+        setup->fov                     = *zoom_fov;
+        g_CHUD->GetSensitivityFactor() = sens_factor;
+    }
+    else
+        g_CHUD->GetSensitivityFactor() = 0.0f;
```
TotallyNotElite, line 45: This looks like a bad idea. We should not set this constantly.

TotallyNotElite (changes requested):

TotallyNotElite (approved):
LGTM

PR #1359, opened 2021-02-19
Improve autobackstab by using the swing trace provided by tf2

BenCat07:
Also adds melee range modifier to melees

### DIFF
--- a/src/hacks/AutoBackstab.cpp
+++ b/src/hacks/AutoBackstab.cpp
@@ -17,6 +17,60 @@ namespace hacks::tf2::autobackstab
 static settings::Boolean enabled("autobackstab.enabled", "false");
 static settings::Int mode("autobackstab.mode", "0");
 
+// Function to do swing trace in a specific direction
+bool doSwingTraceAngle(Vector angles, trace_t &trace)
+{
+    Vector original_angles = LOCAL_E->m_vecAngle();
+
+    re::C_BasePlayer::GetEyeAngles(RAW_ENT(LOCAL_E)) = angles;
+    bool hit_trace                                   = re::C_TFWeaponBaseMelee::DoSwingTrace(RAW_ENT(LOCAL_W), &trace);
+
+    re::C_BasePlayer::GetEyeAngles(RAW_ENT(LOCAL_E)) = original_angles;
+    return hit_trace && (IClientEntity *) trace.m_pEnt != g_IEntityList->GetClientEntity(0);
+}
+
+// Function to do a swing trace against an entity from an arbitary position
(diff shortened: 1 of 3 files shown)


src/hacks/AutoBackstab.cpp
```diff
@@ -17,6 +17,63 @@ namespace hacks::tf2::autobackstab
...
+    Vector original_angles = LOCAL_E->m_vecAngle();
+
+    re::C_BasePlayer::GetEyeAngles(RAW_ENT(LOCAL_E)) = angles;
+
+    bool hit_trace = re::C_TFWeaponBaseMelee::DoSwingTrace(RAW_ENT(LOCAL_W), &trace);
+
+    re::C_BasePlayer::GetEyeAngles(RAW_ENT(LOCAL_E)) = original_angles;
+
+    return hit_trace;
+}
```
TotallyNotElite, line 33: you've
got quite
a few
newlines
here

src/hacks/AutoBackstab.cpp
```diff
@@ -179,63 +236,23 @@ static bool doRageBackstab()
...
-        if (!yangles.empty())
-        {
-            newangle.y = yangles.at(std::floor((float) yangles.size() / 2));
-            current_user_cmd->buttons |= IN_ATTACK;
-            current_user_cmd->viewangles     = newangle;
-            g_pLocalPlayer->bUseSilentAngles = true;
-            *bSendPackets                    = true;
-            return true;
-        }
-    }
```
TotallyNotElite, line 238: Where did this code go

TotallyNotElite (changes requested):

src/hacks/AutoBackstab.cpp
BenCat07: i realized it's actually doing exactly nothing, since looking at the target origin is the best option, and if it does not work looking at the target origin, it just will not work in general, so this code is useless.

src/hacks/AutoBackstab.cpp
BenCat07: i had more code there before which i grouped, but i removed some parts that were unneeded, fixed to be grouped again properly

src/hacks/AutoBackstab.cpp
BenCat07: correction, it does do something, in the case a targets origin is behind a wall, very much an edge case, but i fixed it anyways

aUniqueUser (approved):
Tested on TF2, LGTM

TotallyNotElite (approved):

PR #1364, opened 2021-02-28
Fix crashes with bytepatches on specific hardware due to memory being marked non readable

BenCat07:
Alot of hardware seems to not care about the memory being non readable, but some does, which can cause crashes that cannot really be reproduced. This PR simply marks those memory sections RWX before doing anything to ensure it works.
- [ ] Stable update permission

### DIFF
--- a/include/bytepatch.hpp
+++ b/include/bytepatch.hpp
@@ -44,6 +44,14 @@ public:
         Copy();
     }
 
+    static void mprotectAddr(unsigned addr, int size, int flags)
+    {
+        void *page          = (void *) ((uint64_t) addr & ~0xFFF);
+        void *end_page      = (void *) (((uint64_t)(addr) + size) & ~0xFFF);
+        uintptr_t mprot_len = (uint64_t) end_page - (uint64_t) page + 0xFFF;
+
+        mprotect(page, mprot_len, flags);
+    }
     void Copy()
     {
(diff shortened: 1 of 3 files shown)


include/bytepatch.hpp
```diff
@@ -29,21 +29,39 @@ class BytePatch
...
     }
     BytePatch(void *addr, std::vector<unsigned char> patch) : addr{ addr }, patch_bytes{ patch }
     {
         size = patch.size();
         original.resize(size);
-        memcpy(&original[0], addr, size);
+        Copy();
     }
 
+    static void mprotectAddr(unsigned addr, int size, int flags)
```
BenCat07, line 47: i may want to move this elsewhere

include/bytepatch.hpp
```diff
@@ -29,21 +29,39 @@ class BytePatch
...
+    {
+        void *page          = (void *) ((uint64_t) addr & ~0xFFF);
+        void *end_page      = (void *) (((uint64_t)(addr) + size) & ~0xFFF);
+        uintptr_t mprot_len = (uint64_t) end_page - (uint64_t) page + 0xFFF;
+
+        mprotect(page, mprot_len, flags);
+    }
+    void Copy()
```
BenCat07, line 55: this may want to use mprotectAddr

TotallyNotElite (approved):
Weird flex tf2

PR #1365, opened 2021-02-28
Build -march=x86-64 in binary mode

TotallyNotElite:
This may or may not solve binary mode issues for some users.


BenCat07 (approved):
LGTM

PR #1378, opened 2021-03-18
Aimbot coloring option

LightPower392:
This pull request allows the coloring of the aimbot's target (set in Visuals:Colors) to be disabled for chams and glow.


BenCat07 (approved):
LGTM

Issue #1195, opened 2020-12-14
[Bug Report] Aim at backtrack doesn't work


NotAnUser1:
**Describe the Bug**
Aim at backtrack doesn't aim at backtrack

**To Reproduce**
1. Enable backtrack
2. Enable Fake Latency at any value
3. Enable Aim at backtrack
4. Aimbot doesn't aim at backtrack (result)

**Expected behavior**
Aimbot should aim at backtrack

BenCat07:
aim at backtrack aims at backtrack. besides slow aim maybe.

iraizo:
Related to https://github.com/nullworks/cathook/pull/1381

iraizo:
Related to https://github.com/nullworks/cathook/pull/1381

BenCat07:
this backtrack sucked, new one should fix it

PR #1381, opened 2021-03-22
Rewrite backtrack

BenCat07:
Increases hitrate immensely, also aims to lower code complexity (for modules like aimbot and triggerbot) and does "proper" backtracking, aka teleporting the entity back to where they were originally

### DIFF
--- a/data/menu/nullifiedcat/weapons/aimbot.xml
+++ b/data/menu/nullifiedcat/weapons/aimbot.xml
@@ -123,6 +123,7 @@
         <List width="150">
             <AutoVariable width="fill" target="backtrack.enabled" label="Enable backtracking" tooltip="Allow hitting enemies where they were in the past."/>
             <AutoVariable width="fill" target="aimbot.backtrack" label="Aim at backtracking" tooltip="Aim at backtrack ticks."/>
+            <AutoVariable width="fill" target="aimbot.backtrack.only-last-tick" label="Aim at only the last tick" tooltip="Aim at only the last backtrack tick. !DISABLING THIS MAY CAUSE LAG!"/>
             <AutoVariable width="fill" target="misc.ping-reducer.enable" label="Ping reducer" tooltip="Try to reduce your ping to the number set below."/>
             <AutoVariable width="fill" target="misc.ping-reducer.target" label="Target ping"/>
             <AutoVariable width="fill" target="backtrack.latency" label="Fake latency" tooltip="Amount of fake latency." min="0" max="1000" step="25"/>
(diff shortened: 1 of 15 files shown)


Trapacid:
I tested this and for the first match it seems to work amazingly and flawless with the aimbot aiming at backtrack ticks, a lot better and more accurate than before, but after the second match the game breaks totally with backtracking.

As soon as you pick a class, game locks up and can't seem to do anything until you turn off backtracking entirely. Turning it back on lets you continue playing, but UI elements like map timer, hit damage indicators, control point UI at the bottom, player indicators, kill feed, and map lost or won either don't pop up or they flash repeatedly.

Here are they shown:
https://streamable.com/q0k10p
https://streamable.com/f1uve0

Also, triggerbot no longer works properly being it only aims at the head regardless of the hitbox mode. It seems to be independent of the backtrack enabled option. 
Here's that shown as well:
https://streamable.com/5w3dsh

leKamikaze1:
Gonna try this, bencat is SUPER based

leKamikaze1:
Crashed a ton for me, while pushing the cart, falling from a cliff, or even rocketjumping. Gonna see if these issues are because of the bt_rewrite branch

iraizo:
Closes https://github.com/nullworks/cathook/issues/1195

iraizo:
This branch keeps on crashing when trying to aimbot any kind of building, especially dispensers

BenCat07:
i cannot replivate This PLEASE hand in crash logs.

iraizo:
> i cannot replicate This PLEASE hand in crash logs.

Will do next time im playing because the `report-crash` script tells me its not a cathook crash and i would need to get another log

BenCat07:
use attach-gdb in that case

TotallyNotElite:
What is blocking this PR? Crashes?

leKamikaze1:
ig

BenCat07:
> What is blocking this PR? Crashes?

i fixed a bunch and stopped getting reports

leKamikaze1:
iirc, i sent you another report after what I think are most of the fixes in the telegram chat

leKamikaze1:
nevermind, after my last report, softlock fix was committed

crwn1337:
works great, but aimbot completely breaks if backtrack is disabled (tries to backtrack even if its disabled, models disappearing if you shoot)

BenCat07:
> works great, but aimbot completely breaks if backtrack is disabled (tries to backtrack even if its disabled, models disappearing if you shoot)

Will check, because that does not make sense. Do you mean backtrack itself or "aim at backtrack"?

PR #1383, opened 2021-03-25
Tooltip Improvement

LightPower392:
This pull request fixes tooltips so that they no longer override sliders. Previously the value of a slider could only be shown if there was no tooltip for that menu element. This pull request also adds several new tooltips to the XML files.


BenCat07 (approved):
LGTM

PR #1396, opened 2021-03-30
Extra Autosticky Option + More Tooltips

LightPower392:
This renames the autosticky's legit mode to semi-legit and creates a new legit mode that also requires a line of sight with the enemy, in addition to what semi-legit already does. This also adds more tooltips.

### DIFF
--- a/data/menu/nullifiedcat/antiaim.xml
+++ b/data/menu/nullifiedcat/antiaim.xml
@@ -24,7 +24,7 @@
                     <Option name="Heck" value="11"/>
                 </Select>
             </LabeledObject>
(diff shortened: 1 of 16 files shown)


iraizo (approved):
I found the explaination of the new mode a bit confusing before reading the code of it so im going to try to explain what they do on this fork:
- Rage: detonates regardless if localplayer sees the target/sticky
- Semi-legit: detonates if localplayer sees the sticky
- Legit: detonates if the localplayer sees the sticky and the target at the same time.

I tried it out ingame and they both seem to work also skimmed trough all the tooltips and they seem correct from what I've read, LGTM.

PR #1399, opened 2021-04-01
Add IPC support to autoparty

delimeats-ch:


delimeats-ch:
This doesn't work

delimeats-ch:
Looks like that was all that needed fixing

iraizo (dismissed):
Tested this and seems to work fine all bots joined one specific game, the code also seems fine by taking a quick look of it, LGTM.

delimeats-ch:
The change callback actually won't load hosts from the configured list when IPC mode is disabled since it's not reflected yet so I'm just gonna remove that actually

iraizo (approved):

PR #1415, opened 2021-04-07
Fakelag in air only option

Trapacid:


include/MiscTemporary.hpp
```diff
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <settings/Bool.hpp>
+#include "Registered.hpp"
```
iraizo, line 9: Remove this include

iraizo (changes requested):
After that it looks fine to me.

iraizo (approved):

PR #1416, opened 2021-04-07
Add curl to dependency lists

Trapacid:
Arch uses curl to download gdb binary, submitting crash reports use curl, and migrations use curl. I've found curl to not be installed by default usually.


TotallyNotElite (approved):

iraizo (approved):
Yep, ubuntu as example wont run at all because curl isnt a default package

TotallyNotElite:
Curl is not required for cathook's core functionality.

PR #1417, opened 2021-04-07
Remove strafeprediction for grounded players

BenCat07:
This pull request removes the circle logic applied for air->ground transition.

Also fixes a couple of bugs that make strafe prediction on local player impossible, and uses the proper velocity to determine if a player is moving or not


TotallyNotElite (approved):
Lgtm

PR #1426, opened 2021-04-10
Playerlist and ESP changes

LightPower392:
This pull request makes the playerlist sorted by team. It also changes the ESP by allowing a class icon where the emoji can be, allowing changes in bone thickness, separating corner width from corner height, and fixing performance issues that the emojis and bone ESP were having.
ESP


src/visual/menu/GuiInterface.cpp
```diff
@@ -48,17 +121,20 @@ class PlayerListEventListener : public IGameEventListener
         if (name == "player_connect_client")
```
iraizo, line 121: I know this wasnt in the PR but god damn why isnt this a switch statement

iraizo (changes requested):
Features that work in this pull request:
 - Bone width
 - Playerlist sorting
 - Corner width and height
 - Projectile esp filtering
 - `esp.range` optimization
 
Features that are broken:
- [Sightlines](https://streamable.com/oqga53)
- [Class ESP](https://streamable.com/b8tbpb) (The same issue that light ESP has)
- [Tracers](https://streamable.com/zet4iu)

I have not reviewed the code quality.

LightPower392:
I'll take a look at those

LightPower392:
The tracer found in ESP was really only a repeat of the better version found under the tracers tab, so I merged the two of them. The only tracers are now found in the tab "Tracers"
The glitch with lightESP and the emojis was due to the way their distance size scaling was done, so I replaced that with a different way to scale their size that doesn't cause the glitch. Both are now combined under what is called "Head ESP"
I'm still trying to find out why the sightlines break so often.

LightPower392:
As far as I can tell the issue with the sight lines is due to the game's fog-of-war system. When the sightline disappears, it's not that there was something wrong with the drawing, but rather that the ray trace fails altogether. If there is a fix to this, I'm not aware of it.

TotallyNotElite:
> As far as I can tell the issue with the sight lines is due to the game's fog-of-war system.

No, that's what happens when worldtoscreen returns false.

LightPower392:
I don't think the problem has to do with WoldToScreen. I used `trace.m_pEnt` right after `g_ITrace->TraceRay` to test and found that the problem has to do with the trace ray. It also seems like the sightline only disappears if you look too far away from either the player or what they are looking at. Do you have an idea of what else this might be?

TotallyNotElite:
The behaviour you describe and the behaviour seen in the video perfectly matches what happens with worldtoscreen.

BenCat07:
I am still confused what the issue exactly is. Send a screenshot of an example and i will fix it.

LightPower392:
Irazio posted a video that shows the problem. The sightlines are often not displayed when they should be.

LightPower392:
Any update @BenCat07 ?

BenCat07:
had no time recently

BenCat07:
I am working on this now.

BenCat07:
This is missing GUI updates for the removed rvars, there is alot of red entries in the menu

BenCat07:
Sightlines issue is caused by the marching length being halved, not the whole length of the sightline is being checked for the start position.

BenCat07:
turns out the issue was an `>` being used instead of an `<`

BenCat07 (approved):
I fixed the remaining issues mentioned and tested if there were others.
LGTM.

PR #1441, opened 2021-05-05
Fix problems with targets going out of aimbot range with rapidfire

BenCat07:
Found this out because a user without aimkey sent me his config, i did not think about this possibility.


iraizo (approved):
Looks good to me

PR #1459, opened 2021-05-09
Make Auto-Disguise Select Spy

Lewdcifer:
Auto-Disguise currently only slow us down, so making it select Spy would fix it that :D


BenCat07:
Removes the functionality of autodisguise.

BenCat07 (changes requested):
see my comment

Issue #1442, opened 2021-05-06
Failed to compile cathook error message upon trying to Install Cathook using the GUI


haIfway:
## System Information
**OS:** Ubuntu 20.04.2 LTS x86_64
**DE:** GNOME
**Cathook Version:**  recent version edited byTotallyNotElite on Jul 1, 2020

## Logs / Crash Report  

[CMakeError.log](https://github.com/nullworks/cathook/files/6431009/CMakeError.log)
[CMakeOutput.log](https://github.com/nullworks/cathook/files/6431011/CMakeOutput.log)

## Description  
It repeatedly gives me a "failed to compile" error message upon attempting to use the GUI to install cathook.

**FULL ERROR MESSAGE:**
Cloning into 'cathook'...
~/Desktop/cathook ~/Desktop
From https://github.com/nullworks/cathook
 * [new tag]         stable     -> stable
HEAD is now at 3979a6c Update dependencycheck
Hit:1 http://us.archive.ubuntu.com/ubuntu focal InRelease
Get:2 http://security.ubuntu.com/ubuntu focal-security InRelease [109 kB]
Hit:3 http://us.archive.ubuntu.com/ubuntu focal-updates InRelease
Hit:4 https://apt.kitware.com/ubuntu bionic InRelease
Fetched 109 kB in 1s (85,7 kB/s)
Reading package lists...
Hit:1 http://us.archive.ubuntu.com/ubuntu focal InRelease
Get:2 http://security.ubuntu.com/ubuntu focal-security InRelease [109 kB]
Hit:3 http://us.archive.ubuntu.com/ubuntu focal-updates InRelease
Hit:4 https://apt.kitware.com/ubuntu bionic InRelease
Fetched 109 kB in 1s (95,3 kB/s)
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package libsdl2-dev:i386
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  1 7664k    1 98304    0     0  72602      0  0:01:48  0:00:01  0:01:47 72549
 76 7664k   76 5888k    0     0  2433k      0  0:00:03  0:00:02  0:00:01 2432k
[...]

PR #1460, opened 2021-05-10
Abort if package manager fails

Zanthed:
Aborting if package manager fails should prevent things like #1442 where the package manager fails to install a dependency, yet continues.

Signed-off-by: Ashley <zanthed@riseup.net>


iraizo (approved):
LGTM

I don't know if this fixes the inaccuracy issue but it does remove taunts

BenCat07 (changes requested):
Please remove the "removes taunts" part. I have a more permanent fix coming with a future PR.

Zanthed:
Removed taunt removal.

BenCat07 (changes requested):
There is an issue with this whole thing, if your updater is currently working (system update for example) or some package gets renamed, this whole script Will break completely, in the latter case bricking people's cathook install due to the git update being ran after dependencycheck.

PR #1464, opened 2021-05-12
Fix an issue where doubletap would snap with low FOV

BenCat07:


src/hacks/Aimbot.cpp
```diff
@@ -775,7 +777,7 @@ bool IsTargetStateGood(CachedEntity *entity)
                     {
                         if (ignore_cloak && !(HasCondition<TFCond_OnFire>(entity)) && !(HasCondition<TFCond_CloakFlicker>(entity)))
                             return false;
-                    }
```
BenCat07, line 778: whoever added these spaces, why.

PR #1468, opened 2021-05-13
Fix autobackstab searching for targets even after a valid one was found

BenCat07:
Literally just the title. Break out of the autobackstab loop after finding a valid entity instead of trying to backstab every single one of them.


BenCat07:
This is only a performance increase, it is not involved in other ways.

PR #1469, opened 2021-05-13
Rewrite Setupbones 

BenCat07:
Reconstruct Setupbones on the client so it more closely matches the server. Oh also, add a remove taunts option.


BenCat07:
This is a huge change that probably needs to be tested by people.

crwn1337:
tested with animfix on and seems to work fine without any crashes (i only played with it for an hour)

PR #1476, opened 2021-05-18
fix compile

crwn1337:


crwn1337:
image
****

BenCat07:
The heck, how did that happen? It compiled before

BenCat07:
Actually this is only an issue in your partciular build, as CircleCI compiles it just fine

crwn1337:
it built fine for me before as well i think some update broke it

BenCat07:
Here is what happened: It works fine on normal cathook since everything that includes the file also includes <optional>. The issue is that you have some changes that either Add a file that accesses entitycache.hpp and does not include <optional>, or you cleaned up some includes in an existing file, causing it to error.

No update broke it, as can be seen by the fact that CircleCI compiled cathook successfully on the master branch.

crwn1337:
yes it worked fine for me yesterday until i updated my system, i think some package update broke it is what i meant

PR #1483, opened 2021-06-05
Fix CanHeal() for already healing targets

acidtripparty:
I noticed when the current healing target goes out of sight, CanHeal() returns false.
Add a check to return true if the target is already healing.

### DIFF
--- a/src/hacks/AutoHeal.cpp
+++ b/src/hacks/AutoHeal.cpp
@@ -473,7 +473,7 @@ void UpdateData()
 bool CanHeal(int idx)
 {
     // Already healing, no need to check
-    if(idx == HandleToIDX(CE_INT(LOCAL_W, netvar.m_hHealingTarget)))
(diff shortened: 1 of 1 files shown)


iraizo (approved):
LGTM

acidtripparty:
Can someone merge this?

src/hacks/AutoHeal.cpp
```diff
@@ -472,6 +472,10 @@ void UpdateData()
 
 bool CanHeal(int idx)
 {
+    // Already healing, no need to check
+    if(idx == HandleToIDX(CE_INT(LOCAL_W, netvar.m_hHealingTarget)))
```
BenCat07, line 476: codestyle was not kept, please add a space after the `if`

BenCat07 (dismissed):

BenCat07:
small change requested for codestyle

PR #1485, opened 2021-06-08
Binary search for ProjectilePrediction()

acidtripparty:
Two improvements:

* Use binary search for time prediction in ProjectilePrediction(). We get very tight mindelta with same time complexity. (Flag: `debug.pp-bsearch`)
* Account for projectile gravity drop. (Flag: `debug.pp-gravity`)

Both changes are gated by the flags. No-op code change -- should not affect existing configs.


BenCat07:
Testing today.

BenCat07:
Works just fine

BenCat07:
Actually nevermind, it often just sits there and does not even try to shoot

BenCat07 (changes requested):
Seemingly does not work, see
https://youtu.be/K6lksVIHVkA
(Aim key is being held down once target enters air)

BenCat07:
After 2 years of inactivity, we hope it was worth the wait.

PR #1491, opened 2021-06-22
Fix broken signatures after tf2 update

BenCat07:


BenCat07:
after CI works i will backport to stable

PR #1495, opened 2021-06-23
Fix more crashes

BenCat07:
These were later reported, and fixed after way too much work (for a singular byte)


BenCat07:
Most of this PR is clang format changes to files as the original authors used mixed tabs and spaces and/or not clang formatted files

PR #1496, opened 2021-06-26
Add better Anti-aim options to HvH menu

SheepyChris:
This pull request addresses a multitude of issues and complaints related to the functionality of Anti-aim in cathook. Previously you were not able to independently set fake and real angles for both pitch and yaw angles - this has now been solved with the creation of `antiaim.pitch.fake`, `antiaim.pitch.real`, `antiaim.yaw.fake` and `antiaim.yaw.real` with the additional static/custom values that can be set to each angle. 

image

I've been unable to test if fake yaw angles are working due to them not being visible in local servers (if that is what's happening), so I ask if anyone else could test them and determine that they're good to go for merging. I also apologize if the tab formatting looks different in GitHub, both Featherpad and Geany do that on my end.

### DIFF
--- a/data/menu/nullifiedcat/antiaim.xml
+++ b/data/menu/nullifiedcat/antiaim.xml
@@ -47,11 +47,10 @@
 							<Option name="Spin" value="6"/>
 							<Option name="Edge" value="7"/>
 							<Option name="Sideways" value="8"/>
-							<Option name="Jitter" value="9"/>
-							<Option name="Heck" value="10"/>
-							<Option name="Omega" value="11"/>
-							<Option name="Random" value="12"/>
-							<Option name="Random Clamped" value="13"/>
+							<Option name="Heck" value="9"/>
+							<Option name="Omega" value="10"/>
+							<Option name="Random" value="11"/>
+							<Option name="Random Clamped" value="12"/>
 						</Select>
 					</LabeledObject>
 					<LabeledObject width="fill" label="Real Yaw">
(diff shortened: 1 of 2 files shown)


SheepyChris:
It's been suggested in the Telegram that the switch logic for both yaw modes can be simplified - if the fake portion is confirmed to work I will also add that into the PR.

SheepyChris:
Was able to confirm in game that the fake portion works, if anyone has any better ideas to optimize the switch cases let me know.

EngineerGaming69:
cathook is 5 years late on having good aa

dankdonk:
This is working well for me so far. Also, I don't know about geany but with vim you can just open a file and do
`:set expandtab`
`:set tabstop=4`
`:set shiftwidth=4`
`:retab`
to convert all the tabs to 4 spaces

iraizo:
> 
> 
> This is working well for me so far. Also, I don't know about geany but with vim you can just open a file and do
> `:set expandtab`
> `:set tabstop=4`
> `:set shiftwidth=4`
> `:retab`
> to convert all the tabs to 4 spaces

You can also just use clang format.

BenCat07 (changes requested):
This works pretty well, atleast for the fake yaw part. Fake pitch part also works, but you've accidentally flipped the order, if fake and real pitch are enabled, others will see your real pitch, and your head is at your fake pitch, needs to be fixed.

BenCat07 (approved):
turns out i was just testing it in an unreliable environment.

PR #1503, opened 2021-07-14
Add two small fixes

BenCat07:
Fix an issue with nospread
fix a crash with projectile prediction


EngineerGaming69:
nospread fix 👀

PR #1505, opened 2021-07-15
Add a ton of projectile prediction improvements

BenCat07:
- Huntsman Aimbot now properly aims, works with CanShoot on or off, and with AutoShoot disabled
- Pills are now properly aimed
- Crossbow now targets teammates if they are missing health
- Prediction now no longer passes through walls or gets stuck on janky static prop hitboxes
- Make The Loose cannon work with aimbot and crithack


Snephanie:
poop

PR #1523, opened 2021-08-09
Fix  zoom sensitivity ratio tweak

vicinity-gush8:
They were switched for some reason


vicinity-gush8:
Idk what happened, should be normal now.

Frenzoid (approved):

BenCat07 (approved):
Can confirm, This fixes it

PR #1529, opened 2021-08-16
Fix crashes with radar when object team is -1

d3fc0n6:
only really affects us if we spawn an object via console

### DIFF
--- a/src/hacks/Radar.cpp
+++ b/src/hacks/Radar.cpp
@@ -244,7 +244,7 @@ void Draw()
         ent = ENTITY(i);
         if (CE_INVALID(ent))
             continue;
-        if (ent->m_iTeam() != -1)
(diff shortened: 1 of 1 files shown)


src/hacks/Radar.cpp
```diff
@@ -244,6 +244,8 @@ void Draw()
         ent = ENTITY(i);
         if (CE_INVALID(ent))
             continue;
+        if (ent->m_iTeam() != -1)
```
BenCat07, line 247: You need to check if team is 0, not continue when it's not, this just breaks radar for buildings

BenCat07 (changes requested):

BenCat07 (approved):
LGTM

EngineerGaming69:
a defcon pr????? lets go

PR #1531, opened 2021-08-17
Fix crash with invalid weapons

d3fc0n6:
Only affects us if we have a null weapon
also fixes t-pose crashes


BenCat07 (approved):

src/hacks/Warp.cpp
```diff
@@ -620,6 +620,8 @@ void warpLogic()
         return;
     if (CE_BAD(LOCAL_E) || !LOCAL_E->m_bAlivePlayer())
         return;
+    if (CE_BAD(LOCAL_W)
```
BenCat07, line 623: where bracket

d3fc0n6: funny moment

PR #1532, opened 2021-08-18
Add in startup sounds

d3fc0n6:
meowhook represent


BenCat07:
This needs to have its list loaded from a file or smth, else it's good to go pretty much

d3fc0n6:
not a clean solution

BenCat07:
> not a clean solution

as clean as file loadign gets here

PR #1535, opened 2021-08-18
Fix MVM feature so it actually works

d3fc0n6:
did u even test this


BenCat07:
yeah it broke somewhere along the line due to some changes and i didn't bother

BenCat07:
Wait no this one works just fine

BenCat07:
Tested, this pull request is not needed, it works as-is.

PR #1542, opened 2021-08-23
Update tracers.xml

explowz:
Since the new commit added more tracers, we need to change the tooltip to explain to the user that they can choose from multiple tracer effects and not only the Machina.


BenCat07 (approved):
Didn't notice that, thanks

PR #1578, opened 2021-11-05
Remove references to other games. Fixes

ikeakayke:
Code cleanup. Replaced if with switch case where needed. Some other fixes and improvements


ikeakayke:
CircleCI build failed due to them not having a new enough CMake version.

BenCat07:
This won't work then, it uses Ubuntu LTS which we currently support, so you'd be nuking support for it

iraizo:
Lets not bump up the cmake min requirement version for nothing.

Zanthed:
These are a lot of changes that really should have been split up by individual commits to make it a lot easier to review, debug, and test this.
This is something that is going to need a lot of user testing.

Only things I've really noticed are the compiler options here.

Why are you removing `-march=native`? And using `-O3`?

There's no reason to force `-mavx2`. `-march=native` will automatically detect if a preset processor's CPU type if it supports AVX2 and optimise _**any code that uses AVX2**_. Especially since there are CPU specific optimisations out there and other quirks.

O3 isn't recommended unless carefully tested because it can cause compiler build issues, performance regressions such as excessive loop unrolling, or too much code being generated and making very large or bloated binaries; also causing a performance regression. This also makes debugging unnecessarily more difficult because it can be difficult to determine if a bug is from bad code, compiling with O3 in general, or code incompatible with O3.
There have been plenty of cases where O3 results in reduced math computational speeds as well, something software like this heavily relies on.
Cathook tries to support lots of hardware already.

ikeakayke:
Closed due to finding some bugs. Will come back when they will be fixed.

PR #1581, opened 2021-11-12
Remove references to other games. Optimizations, fixes and some new features

ikeakayke:
Removed references to other games, changed if to switch case where needed, added a few new features, relocated _bunnyhop_ code and renamed it _auto jump_
Removed Ignore Hoovy and Explosion Circles


Zanthed:
Reposted from last PR:

There are so many changes here that this really should have been split up by individual commits to make it a lot easier to review, debug, and test this. This just adds lots of difficulty here.
This is something that is going to need a lot of user testing.

Only things I've really noticed are the compiler options here.

Why are you removing `-march=native`? And using `-O3`?

There's no reason to force `-mavx2`. `-march=native` will automatically detect if a preset processor's CPU type if it supports AVX2 and optimise _**any code that uses AVX2**_. Especially since there are CPU specific optimisations out there and other quirks.

O3 isn't recommended unless carefully tested because it can cause compiler build issues, performance regressions such as excessive loop unrolling, or too much code being generated and making very large or bloated binaries; also causing a performance regression. This also makes debugging unnecessarily more difficult because it can be difficult to determine if a bug is from bad code, compiling with O3 in general, or code incompatible with O3.
There have been plenty of cases where O3 results in reduced math computational speeds as well, something software like this heavily relies on.
Cathook tries to support lots of hardware already.

ikeakayke:
`-march=native` does not always correctly detect if a CPU supports AVX2 or not. Yes, it is a rare occasion that I've only encountered a few times, but it is present. Every CPU made after 2013 supports AVX2. Users with a CPU that does not support AVX2 will need to edit CMakeLists.txt and change all the `-mavx2` arguments to `-march=native -mtune=native`. This information will need to be added to the README.md or to the wiki branch.
We have tested cathook with the `-O3` argument for 2 days now on 3 different computers, and there was no downgrade. As the argument states, the resulted code was more optimized from what we were able to observe. We invite you to test this yourself @Zanthed and write a description of what you were able to observe. If anything seemed to perform better, worse or had no impact at all.

src/crits.cpp
```diff
@@ -950,23 +950,23 @@ void Draw()
...
                         {
 
                             weapon_info info(wep);
-                            std::string crit_string = std::to_string((info.crit_count + 1) * 10.0f - info.crit_attempts - 1);
-                            crit_string += " Shots until Crit! ";
+                            std::string crit_string = std::to_string(int((info.crit_count + 1) * 10.0f - info.crit_attempts - 1));
+                            crit_string += " Shots until crit ";
                             if (info.m_flCritTime + 1.0f >= g_GlobalVars->curtime)
-                                crit_string += std::to_string(info.m_flCritTime + 1.0f - g_GlobalVars->curtime) + "s";
+                                crit_string += std::to_string(int(info.m_flCritTime + 1.0f - g_GlobalVars->curtime)) + "s";
```
BenCat07, line 969: i'd like this to still have some floating point digits, yes currently it has to many, but it shouldn't just have none at all

src/hacks/Aimbot.cpp
```diff
@@ -54,28 +58,29 @@ static settings::Float proj_start_vel{ "aimbot.projectile.initial-velocity", "0"
 
 static settings::Float sticky_autoshoot{ "aimbot.projectile.sticky-autoshoot", "0.5" };
 
-static settings::Boolean aimbot_debug{ "aimbot.debug", "0" };
-settings::Boolean engine_projpred{ "aimbot.debug.engine-pp", "1" };
+static settings::Boolean aimbot_debug{ "aimbot.debug", "false" };
+settings::Boolean engine_projpred{ "aimbot.debug.engine-pp", "true" };
 
-static settings::Boolean auto_spin_up{ "aimbot.auto.spin-up", "0" };
-static settings::Boolean minigun_tapfire{ "aimbot.auto.tapfire", "false" };
```
BenCat07, line 61: Yes, minigun tapfire is a bit buggy with nospread, warp, etc, but it shouldn't just be removed

src/hacks/Aimbot.cpp
```diff
@@ -111,18 +116,19 @@ static void spectatorUpdate()
         break;
         // Disable if being spectated
     case 2:
-        if (g_pLocalPlayer->spectator_state != g_pLocalPlayer->NONE)
+        if (g_pLocalPlayer->spectator_state == g_pLocalPlayer->ANY)
```
BenCat07, line 119: Aimbot will enable when you are spectated in firstperson, but not if in thirdperson

src/hacks/Aimbot.cpp
```diff
@@ -525,15 +529,17 @@ static void CreateMove()
...
         }
-        else if (GetWeaponMode() == weapon_melee)
+
+        if (GetWeaponMode() == weapon_melee)
         {
-            DoAutoshoot();
+            Aim(target_entity);
             if (g_pLocalPlayer->weapon_melee_damage_tick)
-                Aim(target_entity);
+                DoAutoshoot();
```
BenCat07, line 542: Switching the order of these breaks melee aimbot entirely, since it never swings.

src/hacks/Aimbot.cpp
```diff
@@ -544,56 +550,64 @@ static void CreateMove()
...
+                IClientEntity *weapon   = RAW_ENT(LOCAL_W);
+                float time_since_attack = SERVER_TIME - NET_FLOAT(weapon, netvar.flLastFireTime);
+
+                int nBulletsPerShot = GetWeaponData(weapon)->m_nBulletsPerShot;
+                if (nBulletsPerShot >= 1)
+                    nBulletsPerShot = ATTRIB_HOOK_FLOAT(nBulletsPerShot, "mult_bullets_per_shot", weapon, 0x0, true);
+                else
+                    nBulletsPerShot = 1;
+                if ((nBulletsPerShot == 1 && time_since_attack > 1.25f) || (nBulletsPerShot > 1 && time_since_attack > 0.25f))
+                    DoAutoshoot(target_entity);
```
BenCat07, line 572: You cannot just force tapfire behaviour

src/hacks/Aimbot.cpp
```diff
@@ -544,56 +550,64 @@ static void CreateMove()
...
-    else if (!hacks::tf2::warp::in_rapidfire)
+    else if (!hacks::warp::in_rapidfire)
         CreateMove();
 }
 
 #if ENABLE_VISUALS
 bool MouseMoving()
 {
-    if ((g_GlobalVars->curtime - last_mouse_check) < 0.02)
+    if ((SERVER_TIME - last_mouse_check) < 0.02)
```
BenCat07, line 597: This may break some functionality while using warp, which affects player tickbase

BenCat07: Same for anything replacing curtime with SERVER_TIME. There is a reason i adjust curtime in warp

src/hacks/Aimbot.cpp
```diff
@@ -1163,22 +1159,22 @@ void DoAutoshoot(CachedEntity *target_entity)
...
         return;
     }
     else
         begancharge = false;
-    if (g_pLocalPlayer->weapon()->m_iClassID() == CL_CLASS(CTFPipebombLauncher))
+    if (LOCAL_W->m_iClassID() == CL_CLASS(CTFPipebombLauncher))
     {
         float chargebegin = CE_FLOAT(LOCAL_W, netvar.flChargeBeginTime);
-        float chargetime  = g_GlobalVars->curtime - chargebegin;
+        float chargetime  = SERVER_TIME - chargebegin;
```
BenCat07, line 1170: once again, replacing curtime with SERVER_TIME may cause issues due to warp/doubletap, leave it unchanged if possible.

src/hacks/CMakeLists.txt
```diff
@@ -44,11 +43,10 @@ set(files "${CMAKE_CURRENT_LIST_DIR}/AutoJoin.cpp"
 target_sources(cathook PRIVATE ${files})
 
 add_subdirectory(ac)
-if(EnableVisuals)
+if (EnableVisuals)
     set(files "${CMAKE_CURRENT_LIST_DIR}/BulletTracers.cpp"
             "${CMAKE_CURRENT_LIST_DIR}/DominateMark.cpp"
             "${CMAKE_CURRENT_LIST_DIR}/ESP.cpp"
-            "${CMAKE_CURRENT_LIST_DIR}/ExplosionCircles.cpp"
```
BenCat07, line 51: This was actually useful though

src/hacks/Misc.cpp
```diff
@@ -226,176 +250,158 @@ void CreateMove()
...
+            if (flags & (FL_ONGROUND | FL_INWATER))
+                break;
+            if (speed < 2.0f)
                 break;
-            }
-            case 2: // Multidirectional Airstrafe,
-                    // Huge Credits to https://github.com/degeneratehyperbola/NEPS, as their airstrafe
-                    // Apparently just works for tf2
-                    // Also credits to "zelta" for porting it to tf2,
-                    // And "Cyanide" for making it work with cathook.
```
BenCat07, line 259: Don't just remove the credits, i left them there for a reason.

src/hacks/Misc.cpp
```diff
@@ -226,176 +250,158 @@ void CreateMove()
...
 
-                constexpr auto perfectDelta = [](float speed) noexcept
-                {
-                    static auto speedVar = CE_FLOAT(LOCAL_E, netvar.m_flMaxspeed);
-                    static auto airVar   = g_ICvar->FindVar("sv_airaccelerate");
-                    // This is hardcoded for tf2, unless you run sourcemod
-                    static auto wishSpeed = 30.0f;
+            constexpr auto perfectDelta = [](float speed) noexcept
+            {
+                static auto speedVar  = 6062.0f; // Engine limit on 3 axis, see: reddit.com/r/tf2/comments/57hhl4/question_is_there_a_maximum_rocket_jumping/d8t9x82
```
BenCat07, line 310: Needs to be properly tested, game itself does not use that value for calculation in AirMove, but instead m_flMaxspeed, as used before.

src/hacks/Misc.cpp
```diff
@@ -226,176 +250,158 @@ void CreateMove()
...
+        AngleVectors2(VectorToQAngle(direction), &forward);
+        Vector negated_direction = forward * -speed;
+
+        current_user_cmd->forwardmove = negated_direction.x;
+        current_user_cmd->sidemove    = negated_direction.y;
     }
 
-    // HL2DM flashlight spam
-    IF_GAME(IsHL2DM())
+    // Tauntslide needs improvement for movement, but it mostly works
```
BenCat07, line 369: could have just taken the old one then, no? It worked for not just forwards and backwards, but also sideways, unless this one is meant for different kinds of taunts?....

src/hacks/Misc.cpp
```diff
@@ -226,176 +250,158 @@ void CreateMove()
...
+            return;
         }
-        else
-            teammatesPushaway = g_ICvar->FindVar("tf_avoidteammates_pushaway");
-
-        // Ping Reducer
-        if (ping_reducer && !hacks::tf2::antianticheat::enabled)
+        int ping = g_pPlayerResource->GetPing(g_IEngine->GetLocalPlayer());
+        static Timer updateratetimer{};
+        if (updateratetimer.test_and_set(500))
```
BenCat07, line 394: Adding this may cause issues with high latency and a ping not exactly 500ms away, it would bounce more, in theory.

src/hacks/Misc.cpp
```diff
@@ -422,27 +428,55 @@ void Draw()
...
+                {
+                    observermode = "1st";
+                    break;
+                }
+                case OBS_MODE_CHASE:
+                {
+                    observermode = "3rd";
+                    break;
+                }
+                case OBS_MODE_POI:
```
BenCat07, line 463: These are control points, or other "Points of interest" you can switch through, and would not be transmitted to us, as we are not the target

src/hacks/Misc.cpp
```diff
@@ -422,27 +428,55 @@ void Draw()
...
+                {
+                    observermode = "POI";
+                    break;
+                }
+                case OBS_MODE_ROAMING:
```
BenCat07, line 468: Roaming has the same issue as above, it's when you can freely move in spec (e.g. on comm)

BenCat07 (changes requested):

include/playerlist.hpp
```diff
@@ -23,11 +23,13 @@ enum class k_EState
     IPC,
     TEXTMODE,
     CAT,
+    PAZER,
```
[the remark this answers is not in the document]
ikeakayke: Soon to be added: Check chat for Pazer cheating accusation template and set the playing sending that message to PAZER. PAZER will then be treated as RAGE.

include/playerlist.hpp
```diff
@@ -23,11 +23,13 @@ enum class k_EState
...
+    ABUSE,
```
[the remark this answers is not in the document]
ikeakayke: Soon to be added: If betrayal limit is reached, then set to ABUSE, which will be treated as RAGE. The 2 lines 26 and 27 were left from something we weren't able to finish, but we will soon finish and make another pull request.

CMakeLists.txt
```diff
@@ -1,12 +1,8 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.20.4)
```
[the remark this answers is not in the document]
TheCrazyInsanity: > Too high for Ubuntu, which is offically supported.

But does anybody really use ubuntu for this?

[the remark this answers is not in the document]
Zanthed: > > Too high for Ubuntu, which is offically supported.
> 
> But does anybody really use ubuntu for this?

Not sure why whether "anybody uses it" is a concern. You can say the same thing about Gentoo support for Cathook; "Does anybody really use Gentoo for this?" Doesn't make sense.
Plenty of people use Ubuntu and want to use Ubuntu because it's super easy. Cathook officially supports Ubuntu.

zelta01:
https://github.com/MistralDev/Mistral
yoinking unfinished code 😳 
image
image
[thats kinda sus](https://www.youtube.com/watch?v=0bZ0hkiIKt0)

CMakeLists.txt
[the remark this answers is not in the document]
KyappyGR: > > Too high for Ubuntu, which is offically supported.
> 
> But does anybody really use ubuntu for this?

im using mint which is basically the same thing

STG7754:
Aw hell OutOfFOVArrow.cpp got broken xd

CMakeLists.txt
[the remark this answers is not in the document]
hellidox: yes

PR #1615, opened 2022-03-27
Removed redundant request for screenshot of settings, and some other information clarity.

SylveonBottle:
Since the config file is now a requested item, we no longer need to ask the user for just literal screenshot of their menu in order to see what they have enabled.

Config is very useful as there may be other modules that can cause a crash when used in combination with another module. Removed "you can remove this line", making it a requirement.

Combined "drag and drop" information together, since it goes hand-in-hand. Made separate information in it's own separate block.

Please replace "paste symbolize-crash-log here instead" with "paste symbolize-crash-log here as well" if both symbolize-crash-log and logs are requested.

As always, "Allow edits by maintainers" has been specifically enabled. If something is not to your liking, you are fully allowed to modify any text to your own accord instead of asking me to fix something.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -12,15 +12,12 @@ assignees: ''
 **Cathook Version:**  
 
 ## Config used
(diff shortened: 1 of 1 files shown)


BenCat07:
nice work

PR #1631, opened 2022-06-11
Add more ways for bots to get health and ammo

EngineerGaming69:
Resupply lockers could cause bots to try and go in enemy spawns but they do that anyways.

### DIFF
--- a/src/hacks/NavBot.cpp
+++ b/src/hacks/NavBot.cpp
@@ -172,7 +172,7 @@ bool getHealth(bool low_priority = false)
             if (!repath_timer.test_and_set(2000))
                 return true;
         }
-        auto healthpacks = getEntities({ ITEM_HEALTH_SMALL, ITEM_HEALTH_MEDIUM, ITEM_HEALTH_LARGE });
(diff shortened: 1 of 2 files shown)


AsE710:
imagine if bots just low health and low ammo

Navbot just keep path on Enemy spawn

this bad ldea

BenCat07:
I wanted to implement this once, seems to not be feasable with normal means. Enemy lockers do not count, and in some maps lockers change locations based on the stage of the game. Additionally, there is atleast one instance where the locker goee under the floor, the bot would indefinetly path towards it and waste a ton of time. The "proper" and basically only ways to do this without issues would be:
- Process the triggers around the resupply station
or
- Listen for resupply events and figure out which lockers your team can touch, and work on a timer so you do not path to lockers your team has not rouched for a certain amount of time

The latter option is not going to work if noone on your team touches the resupply cabinet, but on maps like 2fort you may accidentally brush it, and avoid the need for outside interference.


The former option is not doable normally, as the server keeps all the triggers on its own end. The only way to make it possible would be to copy the entire game state and simulate the servers end of things.



Former option thus would be preferred, and could be reinforced with some additional methods, e.g. on respawn, mark nearest resupply cabinet as valid (within a certain range to avoid issues with cabinet less spawns).


Also, resupply cabinets don't strictly need to be cabinets, they can be any texture, this cannot be solved without the latter approach, but we'd have to deal with that.

CatholicGod:
I have the fix!

Issue #1634, opened 2022-06-23
cathook crashes on inject


hellidox:
## System Information
**OS: Ubuntu 20.04 LTS**  
**DE: Gnome 3.36.8 w/ x11**  
**Cathook Version: latest dev**  

## Config used

Crashes on inject, no config
## Logs / Crash Report  

[cathook-atrius-144109.log](https://github.com/nullworks/cathook/files/8971139/cathook-atrius-144109.log)
[cathook-atrius-144109-segfault.log](https://github.com/nullworks/cathook/files/8971140/cathook-atrius-144109-segfault.log)

## Description  

crashes on inject
## Steps to reproduce  
- [ ] I am able to consistently reproduce this error

inject

## Checklist  
- [X] I have included my DE, WM, and version of Cathook.
- [X] I have included the steps to reproduce and my logs to the best I am able to share.
- [ ] The Developers are allowed to contact me on [the official telegram](https://t.me/nullworks) for further information if they need it. (Put your username here)

Grimgob:
What you get for cheating pussy
ripbozo

RedFire512:
the fact that you use cathook is just said

just play the game lol

obbyfus:
lets hope it stays that way lmao

darkst0rm223:
> the fact that you use cathook is just said
> 
> just play the game lol

get good get lmaobox :troll:

darkst0rm223:
> What you get for cheating pussy ripbozo

i have you're ip address + i live in you're walls

womblee:
> lets hope it stays that way lmao

legit salt

womblee:
> the fact that you use cathook is just said
> 
> just play the game lol

it's just a game

Dashie17:
Damn that explains why my three can't run anymore rn

Taureon:
> > the fact that you use cathook is just said
> > just play the game lol
> 
> it's just a game

games can't be fun in their intended way if cheaters exist, you know?
i think cheaters are forgetting the reason games exist, which is sad...

StuckInThePast:
Its fun to see stupid ppl like you who wear rose colored glasses and delude yourself into thinking that Valve will suddenly start caring about TF2 all of a sudden when they haven't for 3 years.

aUniqueUser:
fixed with https://github.com/nullworks/cathook/pull/1641

PR #1641, opened 2022-06-24
fix for newest update

racistcop:
fix for https://github.com/nullworks/cathook/issues/1634


aUniqueUser:
not gonna test but looks good enough

PR #1673, opened 2022-06-28
Fix GDB and add additional compiler optimizations

STEVE4git:
GDB has had consistent complaints in Issues due to nullworks forcing you to download a version of GDB from 2021 that breaks system libraries (I can't use thread debugging with his version).
This uses the users GDB version and should be way more reliable.
This adds -Ofast which should have been default anyway. Fast math is what every game engine uses and this should cause zero problems (I've tested it extensively and found nothing).
-> Added fast mode that shoots the first Player it sees (Useful for HVH)
-> Made multipoint far more useful. (It now is so that you toggle on or off and the shrink size is chosen based on the hitbox)


BenCat07:
Merged with fixes

PR #1711, opened 2022-07-24
Decent performance increase due to eliminating duplicate efforts+Using range based for loops

STEVE4git:
->Pretty much the same loop runs 30 times throughout cathook
-> What would happen if we just created a vector of the valid ents to prevent looping through the same 1500 invalid ents 30 times?
->Loops that just check players stayed the same (Due to them already being as efficient as possible)
-> Not too intensive on memory either since it just stores pointers and not actual class objects.

### DIFF
--- a/include/entitycache.hpp
+++ b/include/entitycache.hpp
@@ -219,8 +219,8 @@ namespace entity_cache
 {
 
 // b1g fat array in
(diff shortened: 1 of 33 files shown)


flufferyy:
Why?

PR #1712, opened 2022-07-24
Big performance increase due to getting rid of duplicate efforts

STEVE4git:
-> What if we created a vector from the ents that entity_cache already checks CE_GOOD on
-> Saves literally tens of thousands of loop cycles due to not requiring checking all 2048 ents
-> All features I changed seem to work
-> Range-based for loops FTW

### DIFF
--- a/src/hacks/Tracers.cpp
+++ b/src/hacks/Tracers.cpp
@@ -182,7 +182,7 @@ void draw()
     }
     else
     {
-        for (int i = 1; i = g_IEngine->GetMaxClients(); i++)
+        for (int i = 1; i <= g_IEngine->GetMaxClients(); i++)
         {
             // Get and check player
             auto ent = ENTITY(i);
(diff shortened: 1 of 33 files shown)


src/hacks/Tracers.cpp
```diff
@@ -124,59 +124,118 @@ void draw()
...
+                out.y         = extended.y;
+            }
+            draw::Line(draw::width / 2, draw::height / 2, out.x - draw::width / 2, out.y - draw::height / 2, *color, *line_thickness);
         }
-        else
+    }
+    else
+    {
+        for (int i = 1; i = g_IEngine->GetMaxClients(); i++)
         {
```
BenCat07, line 186:
```suggestion
        for (int i = 1; i <= g_IEngine->GetMaxClients(); i++)
        {
```

BenCat07 (approved):
LGTM

EngineerGaming69:
this might be the best thing ive seen from cathook in a while

PR #1724, opened 2022-09-04
fix hiding of cat_ cvars from server

Cultural-Laugh-2648:
it wasn't checking ones with the prefix

https://github.com/sapphonie/StAC-tf2/blob/bb2dd2b856d049ca35c9084c9e6abace2c60e40b/scripting/stac/stac_cvar_checks.sp#L40


BenCat07:
While cat_ was not blacklisted, changing the prefix would cause issues. Good catch.

BenCat07:
as in, this worked fine unless CON_PREFIX changed

PR #1776, opened 2023-01-02
Projectile dodging

STEVE4git:
Yep. Got the idea from a post on UC and decided to make it better. Works pretty well. The size of the map in entity_cache and the vector skip_these are always very small so they negligibly impact performance. This can be further improved upon by navigating out of the way sooner (Since the majority of the people use bots it can be added in navbot or something). This isn't perfect obviously since it's virtually impossible to move enough to dodge some projectiles (say if you're a heavy and a direct hit is shot a point blank), but from my testing it's very impressive. 
Additional changes:
Dejankified some ESP code, removed unnecessary mutex locks, fixed bone esp, moved code into functions, 
Dejankified some createmove code, moved speedhack to it's own function. No one uses it plus it's long enough to be it's own function
Changed postfix to prefix (It's sometimes faster -- plus it's preferred)  
Changed entity_cache to a more efficient data structure. We only care about idx's that are valid or good. Rarely do maps actually get close to the entity limit of 2048.


STEVE4git:
btw it failed because my PR needs https://github.com/nullworks/source-sdk-2013-headers/pull/2
to be merged

PR #1780, opened 2023-01-07
Player Anim Fix

STEVE4git:
Yep this wasn't checked for CE_GOOD so it led to a nullptr dereference


BenCat07:
Player Anim Fix Fix

PR #1794, opened 2023-02-12
add spamvich

AnonymouseCow:
This dandy new feature spams the sandvich and allows you to eat the sandvich infinitely, however you cannot move (it is sandvich)

idk if anything in cathook does this already but its the thought that counts

it should probably be in extras and not collective


Yushe:
Is this even useful?

BenCat07 (commented):
generally i would.merge this, because why not, but some issues have to be addressed or it will lead to crashes

AnonymouseCow:
> Is this even useful?

no

AnonymouseCow:
i forgor this entire thing so idk if that works or not

Subversionary:
the webedits are real

PR #1834, opened 2023-09-24
Boost fix

Markusip:


BenCat07:
sorry, currently very busy with Uni.


This fails to compile because unordered flat map is missing from system, as in CircleCI, so sadly it does not fix the boost issues.

Any complete fix to the boost issues is welcome, be that upgrading to actual latest boost and fixing libnullnexus, or somehow making cmake actually look at the path we give it for boost headers.

PR #1839, opened 2023-11-22
Fixed the fix me in setup bones + Removed NullNex (Useless)

pixelwarrior99:
I fixed somethings


HutchyBen:
++i makes absolutely zero meaningful difference. why?

PR #1842, opened 2023-12-06
Fix issues with aimbot that caused priority to not properly work

HutchyBen:
Removed cd variable in favour of passing a target around.

Removed vischeck from Aim and added to GetTarget function which replaces IsTargetStateGood serving same purpose but making a target instead.

Run aim after target has been got as calling no longer needed in RetrieveBestTarget (this is what fixes priority issue)


BenCat07:
LGTM

BenCat07:
thank you, and sorry for taking so long, busy anscno time to properly verify everything ingame

